PR #1, opened 2026-06-23
feat: implement project foundation and session engine

akramcodez:
## Summary

This PR implements the foundational infrastructure for `prompt-scrub` and completes Project Foundation & Session Engine.

The goal of this PR is to establish the core architecture, session persistence layer, and development tooling before detector and scrubbing logic are introduced.

### Project Setup

* Added TypeScript project configuration
* Added Jest testing setup
* Added ESLint and Prettier configuration
* Added build, lint, test, and format scripts

### Core Types

* Added shared domain types:

  * `Detector`
  * `Finding`
  * `SessionMap`
  * `ScrubRequest`
  * `ScrubResult`
  * `RehydrateRequest`
  * `RehydrateResult`

### Session Storage

* Implemented cross-platform session persistence
* Added JSON-based storage layer
* Added:

  * `readSessionMap`
  * `writeSessionMap`
  * `deleteSessionMap`
  * `listSessions`

### Session Manager

* Added session lifecycle management
* Added deterministic placeholder generation
* Added placeholder index tracking
* Added save and destroy operations

### ESM Support

* Configured ESM-compatible TypeScript, Jest, and ESLint setup
* Ensured build, lint, and test pipelines work correctly

### Validation

* Build passes
* Lint passes
* Tests pass


will-lamerton:
# PR #1 Review — `feat: implement project foundation and session engine`

**Verdict: Needs changes.**

The code is correct and the tooling works end-to-end (build, lint, test all pass), but there are substantive gaps in correctness and security for a tool whose entire value proposition is "local-first privacy" and "strip identifying content". Specifically: zero unit tests for the only two non-trivial source files, session files are written with the default umask (likely world-readable on most Linux/macOS systems), no atomic write, and a naming inconsistency between the package name and the on-disk config directory. Fix the blockers below and this is ready.
---

## Blockers

### 1. No tests for `session-manager.ts` or `storage.ts`

The only test in the suite is `tests/placeholder.test.ts` and it contains a single `expect(true).toBe(true)`. The two files that contain all of the actual logic (~210 LoC) have 0% coverage. The Jest config (`jest.config.cjs:11–12`) explicitly lists them in `testMatch` and `collectCoverageFrom`, so this isn't an oversight in the runner — it's a missing deliverable on a PR whose stated goal is to "complete the session engine".

At minimum, please add tests covering:

- **SessionManager**: new session, resume existing session, `createPlaceholder` is idempotent on duplicate `originalValue`, `rebuildCategoryCounts` on a preloaded map (e.g. with gaps `["Email_1", "Email_3"]` → next is `Email_4`).
- **storage**: `getConfigDir` per-platform, `readSessionMap` returns `{}` on missing file, `writeSessionMap` creates parent dirs, `listSessions` ignores non-`.json` files, `deleteSessionMap` returns `true` on hit and `false` on miss.

### 2. Session files are written with the process umask

`src/session/storage.ts:71` calls `fs.writeFileSync(filePath, ...)` with no `mode` option. On a typical Linux/macOS system with `umask 022`, the resulting file is `0644` — world-readable. For a tool whose directory is `~/.config/prompt-scrub/sessions/<uuid>.json` and whose contents are the literal `originalValue` of every email/secret/path/URL the user has ever scrubbed (a "scrubbing map"), this is a serious privacy regression that directly contradicts the threat model in `docs/threat-model.md:18`.

Please pass an explicit restrictive mode and ensure the parent directory is also restricted:

```ts
fs.mkdirSync(dirPath, { recursive: true, mode: 0o700 });
fs.writeFileSync(filePath, JSON.stringify(map, null, 2), { mode: 0o600, encoding: 'utf-8' });
```

You may also want to call `process.umask(0o077)` defensively before the write so the mode is enforced even if a future caller forgets.

### 3. Non-atomic writes / silent data loss

`src/session/storage.ts:64–74` calls `fs.writeFileSync` directly, which truncates-then-writes. A crash or power loss mid-write leaves a zero-byte or partial JSON file. `readSessionMap` (`src/session/storage.ts:46–56`) then logs to stderr and silently returns `{}`, and the next `save()` will overwrite the corrupt file with an empty map — losing all the user's session mappings forever. For a privacy tool this means the user may inadvertently re-send original PII to the LLM because the placeholder map was wiped, defeating the purpose of the tool.

Recommended fix: write to `${filePath}.tmp`, `fsync`, then `fs.renameSync` over the target. Additionally, when `readSessionMap` hits a `JSON.parse` error it should rename the corrupt file aside (e.g. `${filePath}.corrupt-${Date.now()}`) rather than silently treating it as an empty map.

### 4. Naming inconsistency: `prompt-scrubber` vs `prompt-scrub`

`package.json:2` declares `"name": "prompt-scrubber"`, but `src/session/storage.ts:14,16,19` writes session files under a `prompt-scrub` directory on all three platforms. Every other reference in the repo is `prompt-scrub`:

- `docs/architecture.md:1,6`
- `docs/cli.md` (all command examples)
- `docs/session-management.md:20–22` (the canonical session path docs)
- `docs/threat-model.md:1,3,21`
- `docs/v1-api.md:12,28`
- `package.json:5` (description)

The on-disk path is the user-visible one and should match the binary/CLI name so users can find their data, and so `npm uninstall` (or future cleanup tooling) can find it. The docs, threat model, and CLI examples all say `prompt-scrub`, so `package.json:name` is the outlier — update it to `prompt-scrub` to match.

### 5. Race in `rebuildCategoryCounts` (placeholder collision)

`src/session/session-manager.ts:30–42`. The pattern `/^([A-Za-z]+)_(\d+)$/` matches the whole key, and the code only checks `index >= counts[category]`, setting `counts[category] = index + 1`. That part is fine in isolation, but the placeholders are stored as `originalValue` (the raw secret) — a placeholder key like `Email_1` may collide with a different category that happens to use the same prefix and number if the session map was ever partially written by a different `SessionManager` instance.

Lower severity than #2 and #3, but worth a regression test that loads a pre-populated map with overlapping prefixes (e.g. `Email_2` and `EMail_2` if you decide to allow that, or `Email_1` and `Email_2` plus a stray `URL_1`) and asserts the next emitted placeholder doesn't collide.

---

## Non-blockers

### 6. PR description overstates scope

The body lists `Detector`, `Finding`, `ScrubRequest`, `ScrubResult`, `RehydrateRequest`, `RehydrateResult` as "added", but `git log --diff-filter=A` shows `src/types/index.ts` was added in commit `cbe85cc` ("initialize project structure…"), not in this PR. The same is true for the empty stubs under `src/cli/`, `src/core/`, `src/detectors/`. Cosmetic, but the PR's claim that it "implements the foundational infrastructure" should match the actual diff — the "Session Storage" and "Session Manager" sections are accurate and should stay.

### 7. `.gitignore` is too thin

Only `node_modules/` is ignored. `dist/` is produced by `npm run build` (`tsc` emits `.js`/`.d.ts`/`.map` under `./dist`) and is not ignored. Please add at least:

```
node_modules/
dist/
coverage/
.env
.DS_Store
```

before someone runs the build and commits the output.

### 8. `getExistingPlaceholder` is O(n) on every call

`src/session/session-manager.ts:62–69` does a reverse scan over `Object.entries(this.map)` for every call. For a long session this is the hot path. Build a parallel `Map<originalValue, placeholder>` in the constructor and keep it in sync inside `createPlaceholder`. Becomes O(1) for the hot path. Out of scope for this PR, but worth filing as a follow-up.

### 9. `getMap()` / `getSessionId()` hand out live internal state

`src/session/session-manager.ts:53,58` return the live internal `map`. Any caller can mutate it and silently desync the on-disk representation. Either return a frozen copy or document the trust boundary. Low risk for a local CLI, not a blocker.

### 10. `destroy()` doesn't surface delete failures

`src/session/session-manager.ts:88–92` clears in-memory state, but if `deleteSessionMap` returns `false` (delete failed) the caller only sees a stderr log. Consider returning a boolean or throwing so the CLI can report the failure to the user.

### 11. `listSessions()` ordering is filesystem-dependent

`src/session/storage.ts:88–95` returns results in `readdirSync` order, which is platform-dependent. Users picking "the most recent session" via a UI will get nondeterministic results. Sort by `mtime` desc before returning.

---

## Suggested approach for tests: switch Jest → Ava

The PR sets up Jest (with ts-jest) as the test runner and ships a single placeholder test. For a small, fast, local-first CLI tool, **Ava is a better fit**: built-in ESM/TS support via `@ava/typescript`, parallel test execution out of the box, simpler test-file syntax, no transformer/babel/ts-jest configuration surface, and a much lighter dependency footprint. As a follow-up to this PR, please replace Jest entirely rather than carrying both runtimes.

Recommended changes:

- Delete `jest.config.cjs`.
- Remove from `package.json` `devDependencies`: `jest`, `ts-jest`, `@types/jest`.
- Add devDeps: `ava`, `@ava/typescript` (`typescript` is already present).
- Replace `package.json` scripts:

  ```json
  "test": "ava",
  "test:watch": "ava --watch",
  "test:coverage": "c8 ava"
  ```

  Optional: `"pretest": "tsc --noEmit"` to keep type-checking in the loop.

- Add an `ava` config block in `package.json`:

  ```json
  "ava": {
    "extensions": { "ts": "module" },
    "nodeArguments": ["--no-warnings"],
    "files": ["tests/**/*.test.ts"]
  }
  ```

- Delete `tests/placeholder.test.ts` and replace it with the `tests/session-manager.test.ts` and `tests/storage.test.ts` files called out in blocker #1.

Notes:

- `@ava/typescript` uses esbuild and doesn't require a build step before running tests, which keeps `npm test` snappy.
- For coverage, `c8` (which Ava's docs recommend) replaces Jest's built-in coverage and integrates with the existing `collectCoverageFrom`-style targets via `--include`.
- Migration is low-risk because the current test suite contains one trivial assertion — nothing of substance to port.

Out of scope for this PR but worth a follow-up issue: after the Ava switch lands, add a CI workflow (`.github/workflows/test.yml`) that runs `npm test` and `npm run build` on PRs.

---

## Follow-ups (out of scope, but worth filing)

- **`getExistingPlaceholder` O(n) → O(1)** via a parallel `Map<originalValue, placeholder>` (see #8).
- **Frozen-copy `getMap()`** or document the trust boundary (see #9).
- **`destroy()` should report success/failure** to its caller (see #10).
- **`listSessions()` sort by `mtime` desc** (see #11).
- **`ScrubOptions.disabledDetectors` / `customDetectors` round-trip test** once detectors are implemented — `src/types/index.ts:30–34` defines the shape but there's no consumer in this PR to exercise it.
- **Move `rebuildCategoryCounts` to a module-private helper** and unit-test it directly. Currently `private` on the class, which is fine, but the testability cost of that visibility choice is real for a function with branching logic.

PR #100, opened 2026-08-14
fix: clarify stateless session map contract

### LINKED ISSUE #87: [Bug] Session Map API contract is unclear regarding stateless mode
## Description

The `scrub()` function in `scrub.ts` is designed to support a stateless mode when a `sessionMap` is provided. However, the API contract is slightly ambiguous. Passing an empty object `{}` correctly triggers stateless mode, but if a caller unintentionally passes `sessionMap: undefined` along with a `sessionId`, it silently falls back to disk mode.

## Environment

- **OS**: Any
- **Install type**: Local library

## Steps to Reproduce

1. Call `scrub({ content: "My email is test@example.com", sessionId: "my-session", sessionMap: undefined })`.
2. Check `~/.config/prompt-scrub/sessions/my-session.json` — it will exist on disk.

## Expected Behavior

The API documentation and TypeScript types should make it unambiguous that:
- `sessionMap: undefined` → disk mode (session written to filesystem)
- `sessionMap: {}` → stateless mode (no disk I/O, map returned in-memory only)

Ideally, a runtime warning or TypeScript overload would surface this distinction at the call site.

## Actual Behavior

Passing `sessionMap: undefined` silently falls back to disk mode with no indication. A developer who intends stateless execution (e.g., in a serverless or read-only filesystem environment) gets unexpected disk I/O with no error or warning.

## Logs/Screenshots

Relevant constructor logic in `src/session/session-manager.ts`

## Additional Context

floze-the-genius:
## Description

Clarifies and enforces the `sessionMap` storage contract for `scrub()`:

- omitting `sessionMap` keeps the normal disk-backed behavior without a warning
- explicitly passing `sessionMap: undefined` still uses disk storage, but now warns instead of failing silently
- passing a map, including `{}`, remains stateless, mutates the caller-owned map in place, and performs no session-file I/O
- the public type comments and API/session documentation now describe those modes and the TypeScript `exactOptionalPropertyTypes` distinction

Fixes #87.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring

## Testing Notes

The new regression test was run before the fix: the explicit-`undefined` warning assertion failed while the other 19 `scrub.test.ts` cases, including the empty-map stateless case, passed.

After the fix:

- `pnpm exec ava tests/core/scrub.test.ts`: 20 passed
- `pnpm run test:all`: 207 AVA tests passed; format, types, lint, knip, and dependency audit passed
- Semgrep was skipped by the repository script because it is not installed in PATH
- Biome emitted one existing informational notice that `biome.json` references schema 2.5.3 while the installed CLI is 2.5.7
- `git diff --check`: passed

## Checklist

### DIFF
--- a/docs/features/session-management.md
+++ b/docs/features/session-management.md
@@ -20,6 +20,15 @@ A session map is a simple key-value dictionary where the key is the placeholder 
 }
 ```
 
+## Storage Modes
+
+The Node.js API chooses storage from the presence of `sessionMap`:
+
+- **Disk-backed:** omit `sessionMap`. A provided `sessionId` is loaded from and saved to disk; without one, `scrub()` generates a new ID.
+- **Stateless:** pass a map, including `{}`. The caller owns the map, it is mutated in place and returned, and no session file is read or written. A `sessionId` may still be supplied as metadata, but it does not enable persistence.
+
+Passing `sessionMap: undefined` is equivalent to omitting it, so it selects disk-backed storage and emits a runtime warning. Pass `{}` when stateless execution is required, especially in serverless or read-only environments.
+
 ## Storage Location
 
 Sessions are stored as JSON files on disk in the user's config directory:
--- a/docs/getting-started/api.md
+++ b/docs/getting-started/api.md
@@ -6,7 +6,7 @@ sidebar_order: 1
 
 # v1 API Reference
 
-The primary interface for `prompt-scrub` is designed to be simple and stateless from the caller's perspective, delegating session persistence to the library.
+The primary interface for `prompt-scrub` supports both disk-backed sessions and caller-owned stateless sessions.
 
 ## Core Functions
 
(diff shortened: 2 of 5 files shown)


akramcodez (commented):
Hey @floze-the-genius, thanks for the solid work here! The core `scrub()` changes and tests look good, but I’d like one thing addressed before approval:

`src/core/rehydrate.ts` has the same `sessionMap: undefined` behavior as `scrub()`: when a `sessionId` is provided, it silently falls back to reading from disk. Since issue is about making this contract unambiguous, could you either apply the same warning/JSDoc treatment to `rehydrate()` or open a follow-up issue for it and link it here?

Once that’s accounted for, I’m happy to take another look :)

will-lamerton (changes requested):
Good, well-scoped fix for #87. `Object.hasOwn` is the right discriminator here, and the tests assert real behaviour (disk file existence, map reference identity) rather than just the warning string. Verified locally on the branch: `tsc --noEmit` clean, `biome check` clean, 20/20 in `tests/core/scrub.test.ts`.

A few things to sort before merge.

**1. `rehydrate()` has the identical ambiguity and no warning**

`rehydrate({ content, sessionId, sessionMap: undefined })` still silently falls back to `readSessionMap(sessionId)` - same bug class, same read-only-filesystem consequence. The new "Storage Modes" section is worded generically ("The Node.js API chooses storage from the presence of `sessionMap`"), so the docs currently over-promise. Either extend the guard to `rehydrate` or scope that sentence to `scrub()` only. Related: `src/core/rehydrate.ts:54` uses truthiness (`reqSessionMap || ...`) rather than presence, worth aligning for symmetry.

**2. Doc wording is now self-contradictory**

`docs/features/session-management.md`: "Passing `sessionMap: undefined` is equivalent to omitting it, so it selects disk-backed storage and emits a runtime warning." After this PR it is not equivalent - omitting is silent. Suggest "behaves like omitting it (disk-backed), but emits a runtime warning."

**3. One type conversion left behind**

`RehydrateRequest.sessionMap` is still `Record<string, string>` in `src/types/index.ts:54` while `docs/getting-started/api.md` now advertises it as `SessionMap`. Convert the third one.

**4. `console.warn` in `src/core` is a new precedent**

Every other `console.*` in the repo lives under `src/cli/`. This is a published library, so embedders with structured logging now get unsolicited console output, and there is no dedupe: a caller looping `scrub()` per message prints once per call. A module-level `let warned = false` guard would make it once-per-process. `process.emitWarning()` is also worth considering since consumers can filter it, though it is Node-only and `console.warn` is safer if a browser build is planned. Either is fine, just want it to be a deliberate call.

**5. Minor: warning string duplicated in the test**

`EXPLICIT_UNDEFINED_SESSION_MAP_WARNING` is not exported, so the test hardcodes the full sentence and a wording tweak fails with an unhelpful diff. Export the constant or assert on a substring. Also `as unknown as Parameters<typeof scrub>[0]` can just be `as ScrubRequest`, ideally with a one-liner noting the cast is needed because our own `exactOptionalPropertyTypes` rejects the literal.

PR #103, opened 2026-08-14
fix: avoid address matches across long prose

### LINKED ISSUE #86: [Bug] `PostalAddressDetector` over-matches on street suffixes in prose
## Description

The regex for `PostalAddressDetector` uses a lazy quantifier `+?` that can match long runs of prose if they happen to end with a street suffix (like "st" or "dr"). 

## Environment

- **OS**: Any
- **Install type**: Global CLI & Local library

## Steps to Reproduce

1. Run the scrub command with `AddressDetector` enabled.
2. Input the text: `There are 42 things you should consider when walking down the st.`
3. Observe the output.

## Expected Behavior

The phrase should not be flagged as a postal address since it doesn't match a standard address structure.

## Actual Behavior

The detector flags the prose as an address because it starts with a number (42) and ends with a street suffix (st).

## Logs/Screenshots

Run this to reproduce:
```bash
echo "There are 42 things you should consider when walking down the st." | prompt-scrub inspect
# Expected: No sensitive entities detected.
# Actual:   [Address]  42 things you should consider when walking down the st.
```

## Additional Context

jonathan-shizuka:
## Description

The postal-address regex can span unlimited prose before a street suffix, so a sentence ending in "st." may be treated as an address. This bounds the street-name segment to one through five words while preserving normal addresses.

Fixes #86

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing Notes

- Targeted Biome formatting and lint checks passed.
- TypeScript typecheck passed.
- A compiled regression probe passed for both the prose false positive and valid address cases.
- The AVA worker cannot start in this Windows sandbox because Node returns `uv_os_get_passwd ENOMEM`; repository CI will run the full suite.

## Checklist

- [ ] If this was for an open issue, I was assigned to it
- [ ] I have formatted, linted, and passed all tests (`pnpm run check`).

### DIFF
--- a/src/detectors/postal-address.ts
+++ b/src/detectors/postal-address.ts
@@ -1,9 +1,9 @@
 import type { Detector, Finding } from '../types/index.js';
 
-// Conservative matching only: number + word(s) + street suffix.
+// Conservative matching only: number + up to five street-name words + street suffix.
 // Examples: 123 Main Street, 45 Oak Road, 1600 Pennsylvania Ave., 10 Downing St, 221B Baker St.
 const ADDRESS_REGEX =
-  /\b\d{1,6}[A-Za-z]?\s+[A-Za-z0-9\s.,'-]+?\b(?:street|st|road|rd|avenue|ave|boulevard|blvd|lane|ln|drive|dr|court|ct|place|pl|square|sq|terrace|ter)\b\.?/gi;
+  /\b\d{1,6}[A-Za-z]?\s+(?:[A-Za-z0-9][A-Za-z0-9.,'-]*\s+){1,5}(?:street|st|road|rd|avenue|ave|boulevard|blvd|lane|ln|drive|dr|court|ct|place|pl|square|sq|terrace|ter)\b\.?/gi;
 
 export class PostalAddressDetector implements Detector {
   readonly name = 'AddressDetector';
--- a/tests/detectors/postal-address.test.ts
+++ b/tests/detectors/postal-address.test.ts
@@ -58,3 +58,10 @@ test('does not fire on plain text', (t) => {
   const findings = detector.detect('I live down the street from here.');
   t.is(findings.length, 0);
 });
+
+test('does not treat long prose ending in a street suffix as an address', (t) => {
+  const findings = detector.detect(
+    'There are 42 things you should consider when walking down the st.',
+  );
+  t.is(findings.length, 0);
+});


akramcodez (commented):
Great fix, thank you! @jonathan-shizuka  The implementation looks solid. Just one small request before merging: could you add a regression test specifically for the 6-word gap?

This would help lock in the `{1,5}` boundary and ensure we don't accidentally expand the regex limit in the future. Something like:

`t.is(detector.detect('The score was 42 a b c d e f street').length, 0);`

Everything else looks good to me!

will-lamerton (changes requested):
Good diagnosis and the right direction, but this currently breaks two existing tests.

### Blocking: greedy quantifier merges adjacent addresses

I ran the full suite on this branch - 2 of 206 tests fail:

```
detectors > postal-address > detects multiple addresses      (expected 2, got 1)
core > scrub > postal address round-tripping works correctly
    - 'Send it to «Address_1»'
    + 'Send it to «Address_2» or «Address_1»'
```

The old pattern used a lazy `+?`, so it stopped at the first street suffix. The new `{1,5}` is greedy, so it tries 5 words first and swallows the gap between two addresses:

```
'From 10 Downing St to 1600 Pennsylvania Ave.'
  before: ["10 Downing St", "1600 Pennsylvania Ave."]
  after:  ["10 Downing St to 1600 Pennsylvania Ave."]
```

That is a correctness issue beyond the test counts: two distinct addresses collapse into one placeholder and the filler between them gets redacted too, which corrupts round-tripping.

One-character fix - make the repetition lazy so it still prefers the nearest suffix while `{1,5}` caps the reach:

```diff
-  ...(?:[A-Za-z0-9][A-Za-z0-9.,'-]*\s+){1,5}(?:street|st|...|ter)\b\.?/gi;
+  ...(?:[A-Za-z0-9][A-Za-z0-9.,'-]*\s+){1,5}?(?:street|st|...|ter)\b\.?/gi;
```

I applied exactly that on your branch and the full suite passes 206/206, including your new prose test.

Please also add a regression test for a wider gap, e.g. `'Ship from 10 Downing St over to 1600 Pennsylvania Ave.'`. The existing multi-address test only caught this incidentally because its filler is a single word.

### Non-blocking

Five words is generous, so shorter prose still false-positives on both the old and new regex:

```
"We reviewed 15 items before heading to the ct."   matches
"The 3 dogs ran down the road."                    matches
```

So #86 is narrowed rather than closed. That is a fine incremental step, but worth softening "while preserving normal addresses" in the description. A stop-word or capitalisation heuristic is the durable fix and can be a follow-up issue.

### Nice

The new inner class excludes whitespace and is delimited by `\s+`, so adjacent repetitions cannot overlap. That removes the backtracking ambiguity the old lazy class had, which is a real win for a scrubber running on untrusted input. Comment on line 3 was updated alongside the regex too.

Happy to merge once the `?` is in and `pnpm run check` is green.

PR #104, opened 2026-08-14
fix(phone): derive span start index from match[0] and value length di…

### LINKED ISSUE #85: [Bug] Phone Detector span calculation is off-by-one for capture groups
## Description
In `phone.ts`, when the regex uses a capture group (`match[1]`), the span is calculated from `match.index`. However, `match.index` points to the start of the *full match* (`match[0]`), not the capture group itself. If the capture group doesn't start at position 0 (e.g., due to lookbehinds), the span will be misaligned.
## Environment
- **OS**: Any
- **Install type**: Local library
## Steps to Reproduce
1. Import and instantiate `PhoneDetector` directly from the library.
2. Call `detector.detect("Call me at +1-555-123-4567 anytime")` and inspect the returned `Finding.span`.
3. Compare `span[0]` against `text.indexOf("+1-555-123-4567")` — they should be equal.
4. Note that the span currently works by coincidence because the lookbehinds are zero-width. Any future change to the pattern that adds a non-zero-width prefix will silently misalign the span.
## Expected Behavior
The span should accurately and robustly reflect the start and end positions of the capture group, even if the regex patterns are modified in the future.
## Actual Behavior
The span calculation uses `match.index` directly instead of deriving the start from the difference in lengths (`match[0].length - value.length`), which is fragile and will produce misaligned spans if the regex patterns ever include a non-zero-width prefix before the capture group.
## Logs/Screenshots
Relevant code in `src/detectors/phone.ts`

## Additional Context

addyCooks:
## Description
In `PhoneDetector`, when regex patterns use capture groups (`match[1]`), using `match.index` directly assumes the capture group starts at index 0 of `match[0]`. If regex patterns ever include a non-zero-width prefix before the capture group, `match.index` would produce misaligned spans. This PR updates the span start calculation to `match.index + (match[0].length - value.length)`, making it robust and consistent with other detectors in the codebase (e.g. `PathDetector`, `URLDetector`, `SecretDetector`).

Fixes #85

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing Notes
- **Automated tests run**:
  - `npx ava tests/detectors/phone.test.ts` (14/14 tests passing)
  - `npx ava tests/detectors/*.test.ts` (95/95 tests passing)
- **Verification**:
  - Added unit test `span accurately reflects value slice in text` to ensure `text.slice(start, end)` strictly equals `finding.value`.

## Checklist
- [ ] I have added/updated documentation if necessary.

### DIFF
--- a/src/detectors/phone.ts
+++ b/src/detectors/phone.ts
@@ -26,9 +26,11 @@ export class PhoneDetector implements Detector {
         const digits = value.replace(/\D/g, '');
         if (digits.length < 7) continue;
 
+        const start = match.index + (match[0].length - value.length);
+
         raw.push({
           category: 'Phone',
-          span: [match.index, match.index + value.length],
+          span: [start, start + value.length],
           value,
           placeholderPrefix: 'Phone',
         });
--- a/tests/detectors/phone.test.ts
+++ b/tests/detectors/phone.test.ts
@@ -86,3 +86,13 @@ test('resolves overlapping matches by picking the longer one', (t) => {
   t.is(findings[0]?.value, '+1-23-4-567');
   t.is(findings[1]?.value, '567-890-1234');
 });
+
+test('span accurately reflects value slice in text', (t) => {
+  const text = 'Call me at +1-555-123-4567 or (555) 123-4567 anytime.';
+  const findings = detector.detect(text);
+  for (const finding of findings) {
+    const [start, end] = finding.span;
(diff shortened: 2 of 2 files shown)


akramcodez (approved):
LGTM @addyCooks :)

PR #107, opened 2026-08-17
feat(cli): add init and config show commands

### LINKED ISSUE #98: [Feature] --config / Config File Init Command
## Description

Add `init` and `config` commands to generate and manage the `~/.config/prompt-scrub/config.json` file.

## Use Case

Currently, users have to manually create the configuration file and guess the schema based on documentation. This is undiscoverable and leads to misconfiguration.

## Proposed Solution

- `prompt-scrub init` generates a default config file.
- `prompt-scrub config show` prints the active configuration.

## Alternatives Considered

Relying solely on CLI flags, which is tedious for complex rule-pack configurations.

## Additional Context

addyCooks:
Closes #98

## Description

The config file schema was undiscoverable users had to hand-create `~/.config/prompt-scrub/config.json` and infer its shape from the docs.

- `prompt-scrub init` writes a default config file (and its parent directories), refusing to clobber an existing one unless `--force` is passed.
- `prompt-scrub config show` prints the active configuration as JSON on stdout and its source path on stderr, falling back to defaults when no file exists.
- Config files are now validated: unknown keys, wrong types, non-string array members, unreadable files and malformed JSON are reported with actionable messages and a non-zero exit, so `config show` doubles as a config check in scripts. Invalid entries are still dropped silently at scrub time, so a broken config never breaks a scrub.
- Corrects the rule-pack docs, which still pointed at the `package.json` `prompt-scrub` key removed in 1.0.1.

`loadConfig()` is now a thin wrapper over the new `readConfigFile()` same inputs, same outputs, no behaviour change for existing callers, and the library still writes nothing to stdout/stderr.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,23 @@ const { content } = rehydrate({
 console.log(content); // "I see your key is sk-12345"
 ```
 
+## Configuration
+
+`prompt-scrub` reads an optional config file for extra rule packs and URL allowlisting. Create one pre-filled with the default schema:
+
+```bash
+prompt-scrub init
+# Created config file at /home/alice/.config/prompt-scrub/config.json
+```
+
+Print the configuration that is actually active — JSON on `stdout`, the path it came from on `stderr`:
+
+```bash
+prompt-scrub config show
+```
+
+Entries that do not match the schema are reported and exit non-zero, so `config show` doubles as a config check in scripts. See the [CLI reference](docs/getting-started/cli.md#configuration) for the full schema and options.
+
 ## Documentation
 
 Full user guides and architecture details are in the [`docs/`](docs/) directory:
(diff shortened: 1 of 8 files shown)


akramcodez (commented):
Great work! @addyCooks The implementation looks solid and the tests/validation are well covered. Just one small request before merging: could you also add a short mention of `prompt-scrub init` and `prompt-scrub config show` to the README for better discoverability?

Everything else looks good to me!

addyCooks:
Thanks for the review, @akramcodez!

Added the README mention in da3d388
short ## Configuration section covering prompt-scrub init and prompt-scrub config show, linking out to docs/getting-started/cli.md for the full schema.

Also fixed a small docs nit in cf84485: 
the config show example showed inline arrays, but the real output expands them one per line.

Both commits are docs only.

PR #112, opened 2026-08-19
feat: encrypt local session files at rest

### LINKED ISSUE #94: [Feature]  Session Encryption at Rest
## Description

Encrypt session JSON files on disk so that the sensitive mappings (e.g., `«Secret_1» -> sk-12345`) are protected.

## Use Case

The threat model acknowledges that session maps are stored in plaintext. If a user's machine is compromised or a laptop is stolen, an attacker can read all previously scrubbed secrets and PII from the `~/.config/prompt-scrub/sessions/` directory.

## Proposed Solution

- Use AES-256-GCM to encrypt the files.
- Support key management via the OS keychain (e.g., using `keytar`), an environment variable (`PROMPT_SCRUB_KEY`), or a manual passphrase prompt.

## Alternatives Considered

Leaving it to full-disk encryption (FDE), which protects against physical theft but not against malware running under the user's account.

## Additional Context

akramcodez:
## Description
Implements AES-256-GCM encryption for local session files at rest to protect sensitive session mappings. Adds an `encryptionEnabled` configuration option to maintain backward compatibility. Encryption keys are securely managed via the `PROMPT_SCRUB_KEY` environment variable or an interactive, muted terminal prompt. Closes #94

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring

## Testing Notes
- **Automated tests run?** Yes, added comprehensive unit and CLI E2E tests for cryptography, storage routines, wrong key failures, data tampering, and headless environment errors. `pnpm run check` executed successfully (213 tests passed, Knip clean).
- **Manual verification steps?** Manually verified the interactive password prompt correctly captures keys natively and via the environment variable in both standard and piped scenarios.

## Checklist


will-lamerton (changes requested):
Thanks for this - the crypto primitives are well chosen (AEAD, per-file random salt and IV, versioned envelope, sane scrypt params, `0600` preserved). The issues are all in the plumbing around them.

## Blocking

**1. Silent downgrade to plaintext.** `writeSessionMap` decides purely on `config.encryptionEnabled`, while `readSessionMap` decrypts based on file contents. So if encryption is later disabled, an already-encrypted session gets rewritten in the clear on the next scrub. `scrub.ts:102` makes it worse by resolving the key when `isSessionEncryptedSync(sessionId)` is true even with encryption off, so the CLI actively unlocks the session in order to write it back unencrypted. The write path should be `config.encryptionEnabled || isSessionEncrypted(id)`.

**2. Decryption failures crash with an unhandled rejection.** Every call site wraps only `getEncryptionKey()` in try/catch and then calls storage outside it (`rehydrate.ts:58`, `scrub.ts:114`, both `sessions` actions). A wrong `PROMPT_SCRUB_KEY` passes key resolution, then `readSessionMap` throws from a place nothing catches - stack trace instead of the clean one-line error the message was written for.

**3. A malformed auth tag quarantines the session.** `decipher.setAuthTag()` in `crypto.ts:63` is outside the try/catch, so `Invalid authentication tag length` escapes unchanged. `readSessionMap`'s new substring filter (`storage.ts:53-61`) doesn't match it, so the file falls through to the corrupt-file branch, gets renamed `.corrupt-<ts>`, and `{}` is returned. Please use a typed error (`class SessionDecryptionError extends Error`) instead of matching on message text, and move `setAuthTag` inside the try.

## Correctness

- `getEncryptionKey` assigns `cachedKey` before validating, so a whitespace-only passphrase throws on the first call but is returned happily on the second. The env var path skips the `trim()` check entirely.
- No passphrase confirmation on first encryption - a typo locks the session permanently with no re-key command. Prompt twice when creating.
- No migration path: enabling `encryptionEnabled` leaves existing sessions plaintext until they happen to be rewritten. A `sessions encrypt` command would close this.
- Node API users can't supply a key at all (module global fed only by env var or TTY). This is a published library, so an exported setter would help.
- `isSessionEncryptedSync`'s comment says "without attempting to parse the full map" but it does exactly that.

## Performance

`deriveKey` runs a full scrypt on every read and every write with no cache. `sessions list` does `isSessionEncryptedSync` (read + parse) plus `readSessionMap` (read + parse + KDF) per session, so listing 20 encrypted sessions takes seconds. Cache derived keys by salt.

## Tests

- The description says CLI E2E tests were added, but no file under `tests/cli/` is touched. The whole CLI layer, where issues 1-3 live, is untested.
- The two new storage tests are racy: AVA runs tests in a file concurrently, and both mutate a process-wide global config plus `PROMPT_SCRUB_KEY`. Existing tests in that file will intermittently hit "no key is available". They need `test.serial`.
- `process.env.PROMPT_SCRUB_KEY = originalEnv` sets the literal string `"undefined"` when the var was unset - use `delete`. The second test never restores the env var and relies on the first test having created the config dir, so it fails in isolation.
- `t.false(Object.keys(parsed).includes('«Secret_1»'))` only checks top-level keys. Prefer `t.false(rawData.includes('sk-1234'))`.

## Style and docs

- `err: any` in five places; the surrounding code uses `err: unknown` with a cast. `encryptSession(data: any)` should be `SessionMap`. `catch (err: any)` in `crypto.ts:70` never uses `err`.
- The `process.exit(1); return;` block is copy-pasted four times across three files - a `resolveKeyOrExit()` helper would collapse it and give one place to also fix issue 2.
- `docs/getting-started/cli.md` documents the config file but wasn't updated with `encryptionEnabled` or `PROMPT_SCRUB_KEY`.
- README should note that `rehydrate` on an encrypted session also needs the key, and that the inline `PROMPT_SCRUB_KEY=...` example puts the passphrase in shell history.

Requesting changes mainly for 1-3 plus the missing CLI tests. The core crypto is solid, it's the integration that needs another pass.

will-lamerton (changes requested):
The AES-256-GCM envelope itself is solid: random per-file salt and IV, tag verification, no rolled-own primitives, good crypto unit tests. The problems are at the integration seams, plus some unrelated changes bundled in.

## Blocking

1. **`watch` never resolves a key.** `watch` -> `handleScrub` -> `scrub()` -> `session.save()` -> `writeSessionMap()`. With `encryptionEnabled: true` and no `PROMPT_SCRUB_KEY`, this throws `SessionDecryptionError` mid-loop, uncaught, with no TTY prompt offered. Needs `resolveEncryptionKeyOrExit()` before the watch loop starts.

2. **`scrub` still prints a raw stack trace on a wrong key.** `src/cli/commands/scrub.ts:143` calls `handleScrub()` outside any try/catch, so `scrub --session-id <encrypted-id>` with the wrong key propagates to `src/cli/index.ts:50` and `console.error(err)` dumps the whole Error. `rehydrate.ts` and `sessions.ts` got the try/catch; `scrub.ts` didn't. This is the exact failure `cli-key-resolver.ts`'s docstring claims to have fixed.

3. **`sessions encrypt <bogus-id>` fabricates a session.** `readSessionMap` returns `{}` for a missing file, then `writeSessionMap` creates an encrypted empty session and reports `Encrypted 1 session(s).` Guard on file existence.

4. **`listSessions` rename is breaking and wrong.** `lastModifiedAt` -> `createdAt` still holds `stats.mtime`, is exposed publicly via `SessionManager.listAll()` (breaking, under a `minor` changeset), and `sessions list` still prints the header `Last Modified`. Please revert; unrelated to encryption.

5. **`init` won't emit `encryptionEnabled`, contradicting this PR's docs.** `createDefaultConfig()` was narrowed with `Omit<..., 'encryptionEnabled'>`, but `docs/getting-started/cli.md` documents the generated file as containing `"encryptionEnabled": false`. The `Omit` also forced `CONFIG_KEYS` into a hardcoded literal, losing the type/validation link. Fix:

```ts
export function createDefaultConfig(): Required<PromptScrubConfig> {
  return { rulePacks: [], urlAllowlist: [], sessionTtlDays: 7, encryptionEnabled: false };
}
const CONFIG_KEYS = Object.keys(createDefaultConfig());
```

## Please split out

- `.semgrepignore` + `semgrep-disable-line` + `nosemgrep` — disabling the scanner across 5 workflows, `code-tell.ts`, and `secret.test.ts` is a security-posture change riding along in a security PR.
- `pnpm-workspace.yaml` supply-chain policy (`minimumReleaseAge: 10080` will start blocking fresh dependency releases in CI).
- README deletes the entire "CLI: Watch clipboard" section to make room for the encryption docs. Add, don't replace.

## Smaller notes

- **scrypt params aren't in the envelope.** `N: 16384, r: 8, p: 1` is hardcoded, so raising the cost later makes every existing file undecryptable. Store `{N, r, p}` in the envelope.
- **No re-key path** — `sessions encrypt` skips already-encrypted files, so passphrase rotation is impossible. Support or document it.
- **`clearCachedEncryptionKey()` doesn't clear `derivedKeyCache`** — derived keys stay resident after the passphrase is cleared.
- **`getEncryptionKey` skips `normaliseInputKey` on the env path**, so `PROMPT_SCRUB_KEY="   "` is accepted while the same value at the prompt is rejected.
- **`normaliseInputKey` doesn't normalise** — it returns input verbatim. Rename to `assertValidKey`, or trim.
- **`decryptSession` reports `JSON.parse` failures as "wrong key"** — GCM already authenticated the plaintext, so a parse failure is a different bug. Move the parse outside the try.
- **`getEncryptionKey({ confirm: true })` is never called** by any CLI path, so the typo protection it exists for doesn't ship. Wire it into the first encrypted write or drop it.
- **`encryptWithKey(map, _confirmed)`** takes an unused param.
- **`cli-key-resolver.ts`** imports `'../core/key-manager.js'` from inside `src/core/` (should be `./`), and its docstring references "review feedback #2".
- **Over-eager prompts** — `sessions list`/`show` demand a key whenever `encryptionEnabled` is set, even if every session on disk is plaintext. Gate on `isSessionEncrypted` alone for read-only commands.
- **Perf** — `writeSessionMap` now does a full `loadConfig()` parse plus a full `isSessionEncrypted()` read+parse on every write; `sessions list` reads each file twice. Fine one-shot, but watch mode does this every poll tick. Also `isSessionEncrypted`'s "Quickly tells the caller" docstring doesn't match a full read+parse.
- Missing newline at EOF in the changeset.

## Tests

Good breadth on the primitives. Gaps: no test for watch + encryption (1), for `scrub --session-id` with a wrong key (2), for `sessions encrypt <nonexistent-id>` (3), or for `sessions show` under encryption. Also:

- `normaliseInputKey returns trimmed key` asserts the value is *not* trimmed (`t.is(normaliseInputKey('  valid-key  '), '  valid-key  ')`). Name or behaviour is wrong, and it matters for pasted passphrases.
- Both TTY tests degrade to `t.pass()` under a TTY, so they cover nothing when run locally in a terminal.
- `null as any` — prefer `as unknown as string`.

## Security notes (non-blocking)

- Worth documenting that `PROMPT_SCRUB_KEY` is readable via `/proc/<pid>/environ` on Linux, not just that inlining it hits shell history.
- Worth stating the threat-model boundary explicitly: encryption protects the file at rest, not the map in memory or `sessions show` output.

PR #113, opened 2026-08-19
feat(cli): add scrub statistics / summary output

### LINKED ISSUE #97: [Feature] Scrub Statistics / Summary Output
## Description

Print a brief summary to `stderr` detailing how many entities were scrubbed after a command completes.

## Use Case

When running `scrub`, the user currently receives the scrubbed text and a session ID, but has no quick way to know what was actually removed without running `inspect` separately.

## Proposed Solution

Print a concise summary to `stderr`:
`Scrubbed: 3 entities (1 Email, 2 Secrets)`

Add a `-q` or `--quiet` flag to suppress this output for strict pipeline usage.

## Alternatives Considered

Running `inspect` before `scrub` to get the finding count — but this requires two separate commands, doubles the processing time, and still doesn't surface a summary after the actual scrub operation. Having the summary appear automatically as part of `scrub`'s stderr output is much lower friction and doesn't require any change to existing scripts (which can simply redirect stderr if they don't want it).

## Additional Context

addyCooks:
Closes #97.

`scrub` now reports what it replaced on stderr, e.g. `Scrubbed: 3 entities (1 Email, 2 Secrets)`, so the counts no longer require a separate `inspect` run. `-q`/`--quiet` suppresses it for strict pipeline usage; the `Session ID:` line is left alone since scripts need it to rehydrate.

`scrub()` exposes the same data to library callers as `result.stats` (`totalEntities`, `byCategory`), counted per replacement and aggregated across every message of a `Message[]` input.

## Description

`scrub` previously gave you the scrubbed text and a session ID, but no way to know *what* was removed without a second `inspect` run doubling the processing and still not describing the actual scrub. This adds a one-line summary on stderr:

```console
$ echo "Mail alice@acme.com about sk-abcdefghijklmnopqrstuvwxyz" | prompt-scrub scrub
Mail «Email_1» about «Secret_1»
Session ID: 6f1c2b90-0d3a-4f8e-9a21-2b7c1e4d5a63
Scrubbed: 2 entities (1 Email, 1 Secret)
```

Implementation notes:
- `ScrubStats` (`totalEntities`, `byCategory`) is added to `ScrubResult`, so the data is available to library callers too no CLI-only logic.
- Counting happens where findings are already resolved in `scrubString`, so there's no second detector pass.
- Counts are collected left-to-right (replacement stays right-to-left), so `byCategory` reads in the order entities appear in the text.
- Summary goes to stderr, keeping stdout byte-identical for pipes.

Three decisions worth a look:
- **`--quiet` suppresses the summary only.** The `Session ID:` line still prints silencing it would leave scripts unable to rehydrate. Happy to change if you'd rather `-q` mean "no stderr at all".
- **Counts are replacements, not unique values.** A value appearing 3× counts 3×, even though all three collapse onto one placeholder this matches what `inspect` lists. Reusing `--session-id` never carries earlier counts over; the summary always describes the current run.
- **Zero case prints `Scrubbed: 0 entities`** rather than staying silent, so "nothing found" is distinguishable from "flag forgotten".

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Checklist


akramcodez:
So great work on it @addyCooks, merging it :)

PR #114, opened 2026-08-20
feat(detectors): add locale-aware detector support

### LINKED ISSUE #96: [Feature] Multi-Language / i18n Detector Support
## Description

Introduce support for locale-specific PII detection (e.g., German street formats, Brazilian CPF numbers, Japanese postal codes).

## Use Case

The current detectors (especially `PostalAddressDetector` and `NameDetector`) are heavily biased towards English and US/UK formats. Non-English users receive significantly worse protection.

## Proposed Solution

Create a locale-aware configuration (`prompt-scrub scrub --locale de-DE`). Distribute locale-specific rules as standalone rule-packs (e.g., `@nanocollective/prompt-scrub-locale-de`) that plug into the core engine.

## Alternatives Considered

Hardcoding all global formats into the core tool, which would bloat the binary and slow down regex execution.

## Additional Context

addyCooks:
Closes #96

## Description

Adds the locale seam from #96. This is the core mechanism only  no locale rules are hardcoded, per the issue's Alternatives Considered. `docs/features/authoring-rule-packs.md` gains a Locale-Scoped Rule Packs section with a German example, so `@nanocollective/prompt-scrub-locale-*` packs can be built against a documented contract.

Fully backwards compatible: `locales` is optional, existing rule packs and configs are unaffected, and `rules list` output is byte-identical when no locale detectors are loaded.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring

## Checklist


akramcodez:
@addyCooks Thanks for the PR! Could you please resolve the merge conflicts with `main` so we can get this reviewed?
Also, we recently migrated to Changesets please run `pnpm changeset` to generate a changeset for this new feature and commit the resulting file.

addyCooks:
Hey @akramcodez  ! 
Both sorted,

Conflicts: rebased onto latest main and force-pushed, still a single commit
Changeset: added .changeset/i18n-locale-detectors.md (minor). Missed the migration on my end, good catch 

Quick heads-up: the failing changeset-check isn't coming from this PR. actions/checkout@v7 now refuses to check out fork code in a pull_request_target workflow, so it dies at the checkout step before the actual check runs same on #111, 
and it only passes for dependabot since those branches live in this repo. Opened #121 with a fix. 
This PR isn't blocked by it.

will-lamerton (changes requested):
Nice work - the mechanism-only scope is right, backwards compatibility holds, and the test coverage is above the bar for this repo. A few things to resolve before merge.

### 1. Locale precedence can reduce redaction coverage

`LOCALE_PRECEDENCE_BOOST` makes a locale finding beat an equal-category finding unconditionally, including when its span is strictly narrower. The new test in `tests/core/scrub.test.ts` encodes this:

```
'Adresse: 10 Downing St, Berlin'
  without locale -> 'Adresse: «Address_1», Berlin'      // "10 Downing St" redacted
  with locale    -> 'Adresse: 10 «Address_1», Berlin'   // "10" now leaks
```

Silently narrowing a redaction is the dangerous direction for this tool, and the test currently locks it in as intended behaviour. The goal (letting a locale pack replace an English-biased match) still works if the tie-break refuses to shrink coverage - either only apply the boost when the locale finding is not a strict subset of the competitor, or accept the union of the two spans.

### 2. Unvalidated `locales` from third-party packs can crash the CLI

`src/core/rule-packs.ts` deliberately fail-soft validates `name` and `detect`, but copies `locales` through unchecked. A pack shipping `locales: 'de-DE'` (string instead of array) passes the `.length > 0` guard and then throws:

- `matchesLocale(d.locales, ...)` -> `supported.some is not a function`, so `scrub` throws
- `localesOf(d)` in `rules.ts` -> `.join` on a string

Please validate with `Array.isArray(...)` plus a per-item string check at the loader and drop the field otherwise, consistent with the surrounding style.

### 3. `--locale` is unvalidated and fails silently

`config.locale` is validated and exits 1 on a bad tag, but `options.locale?.trim() || config.locale` accepts anything. `--locale German!`, `--locale de_DEE`, or a valid tag with no matching pack installed all behave identically: no locale detectors run, no message. The user thinks they scrubbed German PII and did not. Suggest validating the flag against `LOCALE_PATTERN`, and warning on stderr when an active locale activates zero detectors.

### Follow-ups (happy for these to be separate)

- `watch` has no `--locale` flag. `WatchStepOptions` mirrors `--disable`, `--strict-name`, `--url-allowlist` but omits `locale`. Configured `locale` still applies via the `handleScrub` fallback so nothing is broken, but watch is the mode where a per-run locale matters most.
- `rules list` resolves state against config only, so it cannot preview what `scrub --locale pt-BR` would enable. Also `defaultState` now carries two meanings: a default for built-ins, an effective state for locale detectors. Consider a `--locale` option and splitting the column.

### Nits

- The `0.5` offset works but is implicit against an integer priority table. A named comparison step or an internal flag on `Finding` would read more clearly.
- Threading preference as a `Set<Finding>` keyed on object identity is fragile - any future code that maps or clones findings between detection and resolution silently loses the preference, with no type-level signal. Low risk today since `detectFindings` owns both ends.
- `src/core/detectors.ts` gains two more mid-file imports. Pre-existing pattern, but a good moment to hoist all three.

Good extras worth calling out: `detectFindings` removing the duplicated detect+resolve logic, the schema-driven `validateConfig` refactor preserving existing error messages, and the CLI e2e tests that write a real rule pack rather than mocking.

addyCooks:
Thanks for the review @will-lamerton 
all three blocking items are done, 
plus both follow-ups and the three nits. 

Pushed as a second commit so the delta is reviewable on its own.

**1.** The tie-break now applies only within an equal-priority collision, and is refused when the locale finding is a strict subspan of the one it would displace. Your example is now identical with and without a locale `'Adresse: «Address_1», Berlin'` both ways, no `10 ` leak. Equal-or-wider locale findings still win; `SecretDetector` still outranks. The test that locked in the old behaviour is replaced by three.

**2.** `locales` is validated at the loader with `Array.isArray` + a per-item string check, and dropped otherwise. Sanitised detectors delegate rather than being spread, so class-based packs keep their prototype `detect()`.

**3.** `--locale` is validated against the same `LOCALE_PATTERN` the config uses on `scrub`/`inspect`/`watch`/`rules list` (exit 1), and a valid tag activating zero detectors warns on stderr. Note `de_DEE` is *accepted* `LOCALE_PATTERN` allows underscores and 2–8 char subtags, and I reused it deliberately so the flag and config key can't disagree; that case is caught by the new warning instead.

**Follow-ups:** `watch --locale` (resolved before the poll loop starts), and `rules list --locale` with the column split into `Default State` / `Locales` / `Locale State`.

---

**On the failing checks they aren't from this PR.**

I rebased onto latest `main`, which clears the CodeQL *"1 configuration not found"* warning: the branch was 10 commits behind and still had the old `ci.yml` that #130 replaced.

Semgrep is separate and pre-existing. All 21 findings are in files this PR never touches, and they're byte-identical to `main`:

```bash
git diff --quiet upstream/main HEAD -- .github/ pnpm-workspace.yaml src/detectors/code-tell.ts
# clean → main fails semgrep the same way

**Review by will-lamerton — CHANGES_REQUESTED** [review]
Nice work - the seam is clean, `Default State` vs `Locale State` is the right split, and the delegating-object trick in `withValidatedLocales` (keeping prototype `detect` on class-based detectors) is a thoughtful catch. I built the branch and ran the suite: 334 passing.

One blocker, plus some smaller follow-ups.

## 1. Locale precedence can shrink redaction coverage (blocker)

`redactsStrictlyLessThan` only guards **strict containment**. On a partial overlap the locale finding wins even when it covers less text and starts later, exposing the leading region that was previously redacted. Verified against the built branch:

```
generic  span [0,20], value length 20
locale   span [15,30], value length 15  (localeScoped)

this branch:  resolveCollisions -> [[15,30]]   chars 0-15 now in the clear
before:       longest-value tie-break -> [0,20]
```

Both input orderings agree, so it is not an ordering fluke. This is a coverage regression versus the old longest-value rule.

Minimal fix in `src/core/collision-resolver.ts` - invert the guard from "not strictly contained" to "must cover":

```ts
function covers(outer: Finding, inner: Finding): boolean {
  return outer.span[0] <= inner.span[0] && outer.span[1] >= inner.span[1];
}

// in candidateWins:
if (covers(preferred, other)) {
  return preferred === candidate;
}
```

The locale finding then only wins when it redacts a superset; everything else falls through to the existing longest-value tie-break. I applied this locally and all 334 tests still pass, including all eight new locale collision tests. Please add the partial-overlap case as a regression test - nothing currently exercises a non-containing overlap.

## 2. "Same category" is really "same priority bucket"

`priorityOf` maps category to detector name with `?? 99`, so every unrecognised category collapses into one bucket. A locale-scoped `Cpf` finding and an unrelated custom `Ticket` finding are "equal priority" and enter the locale tie-break together. Either compare `finding.category` explicitly in the locale branch, or reword `docs/features/detectors.md` and `authoring-rule-packs.md` to "of equal priority".

## 3. Changeset overclaims

> except where that would redact less text than the finding it displaces - a narrower locale match never shrinks an existing redaction

Not true for partial overlaps, per #1. `authoring-rule-packs.md` is accurate ("sits strictly inside"); the changeset and `detectors.md` are not. Fixing #1 makes the claim true and leaves just a wording pass.

Also: our other changesets are one-liners. This one is ~250 words in a single paragraph and lands verbatim in the changelog - please trim it to the user-visible surface, and mention that `getActiveDetectors` now filters custom detectors carrying a `locales` array (previously that property was ignored and the detector ran).

## 4. Test cleanup is not exception-safe

`cleanup()` in `tests/core/rule-packs.test.ts` and `clearWatchConfig()` in `tests/cli/watch.test.ts` run on the happy path only. A failing assertion leaves `PROMPT_SCRUB_CONFIG_DIR` set and a `config.json` in the shared `tmpDir`, cascading into unrelated serial tests. Use `t.teardown(cleanup)`.

## Nits

- `resetLocaleWarning` exists only for tests; `t.teardown` plus the injected warn function would avoid the exported reset.
- `rules list --locale` is a silent no-op when no locale detector is loaded, which is exactly when someone runs it to check whether their pack installed. Consider always printing the active locale when `--locale` is passed explicitly.
- `LOCALE_PATTERN` accepts `de-de-de-de` and rejects private-use `x-foo`. Fine as a gate, just not strict BCP-47.

No new attack surface otherwise - rule packs already execute imported code, and `locales` is validated at the boundary like `name` and `detect`. #1 is the only privacy-relevant item, which is why it blocks.

addyCooks:
Thanks @will-lamerton, 
fixed in f806def:

1. **Locale precedence on partial overlap.** Replaced the strict-containment guard with `covers()` the locale finding only wins when its span is a superset of the one it displaces. Verified against your exact repro and its reverse ordering; added both directions plus the "wins when it covers more" case as regression tests.
2. **Same-category vs same-bucket.** Added an explicit `category` match alongside the priority check, so two different categories sharing the `?? 99` bucket no longer enter the locale tie-break together.
3. **Changeset.** Trimmed to one paragraph, reworded the "strictly inside" overclaim now that #1 covers partial overlap too, and added the `getActiveDetectors` filtering note.
4. **Test cleanup.** Both `cleanup()` and `clearWatchConfig()` now run via `t.teardown`.

Nits taken: `rules list --locale` always echoes the active locale when passed explicitly, and `resetLocaleWarning` tests use `t.teardown` instead of manual resets. Left `LOCALE_PATTERN`'s BCP-47 looseness alone as you suggested.

Also merged `upstream/main` branch was 16 commits behind and dragging in the same stale Semgrep findings as last time; clean merge, dropped straight to the 6-finding baseline.

akramcodez (changes requested):
@addyCooks I found one remaining blocker in the locale collision handling in `src/core/collision-resolver.ts` (`candidateWins`).

When a locale-scoped finding partially overlaps a same-category generic finding but does not fully cover it, the current logic falls through to the normal `value.length` tie-break. This means a longer locale finding can still win and shrink the redacted span, potentially leaving part of the original match exposed.

The locale branch should only grant precedence when `covers(preferred, other)` is true. Otherwise, the locale finding should lose regardless of value length:

```ts
if (covers(preferred, other)) {
  return preferred === candidate;
}
if (candidate.localeScoped) return false;
return true; // existing non-locale finding keeps its slot
```

Please also add a regression test for the case where the locale finding partially overlaps, covers less, but has a longer value than the generic finding. For example, generic `[0,20]` with value length 20 versus locale `[15,45]` with value length 30.

The existing test does not catch this because the locale value happens to be shorter, so the normal `value.length` fallback produces the correct result by coincidence.

Once this is fixed and the regression test is passing, I think the locale work should be good to go.

addyCooks:
Thanks @akramcodez, fixed in 4d01b0f

The locale branch now only grants precedence via `covers(preferred, other)`; on a partial, non-covering overlap it loses outright instead of falling through to the `value.length` tie-break:

if (covers(preferred, other)) {
  return preferred === candidate;
}
return !candidate.localeScoped;

Added the regression case you asked for, generic `[0,20]` (len 20) vs. locale `[15,45]` (len 30), plus reran the full locale/collision suite (28 tests) to confirm nothing else regressed.

Also merged `upstream/main` to resolve the conflicts, it had picked up the confidence-scoring feature (#115) since my last push, which touched the same files (`collision-resolver.ts`, `scrub.ts`, `config.ts`), so this update folds locale-scoping and confidence-scoring into one detection pipeline.

akramcodez:
thanks for the PR @addyCooks

PR #121, opened 2026-08-25
ci(changeset-check): read PR files via API instead of checking out fork code

addyCooks:
### Problem

`changeset-check` currently fails on **every** pull request from a fork, including this one. It dies at the first step:

> Refusing to check out fork pull request code from a `pull_request_target` workflow.

`actions/checkout@v7` now blocks this by default (`allow-unsafe-pr-checkout: false`). The job runs as `pull_request_target`, so it carries the base repo's token and secrets pulling fork-authored code into that trusted context is the classic "pwn request" pattern, and checkout refuses it.

It shows up on #114 and #111, and passes only for dependabot, whose branches live in this repo rather than a fork.

Two knock-on effects: the check is documented as a non-blocking nudge that "never fails the build", but it now reports a red X on every outside contribution; and because it dies before its own logic runs, the changeset reminder never actually fires.

### Fix

The job never needed the fork's code. It checked out the full repository only to run `git diff --name-only --diff-filter=A` and look for an added `.changeset/*.md` a question about **filenames**. `pulls.listFiles` answers that from the API, so the checkout can go entirely.

That keeps `pull_request_target` (still required for `pull-requests: write` to post the comment) while removing fork content from the trusted context, rather than opting back into the risk with `allow-unsafe-pr-checkout: true`.

### Notes

- Detection rule is unchanged: added-only, `.changeset/*.md`, `README.md` excluded. Verified against the cases that matter — `config.json`/`changelog.cjs` still ignored as non-`.md`, `modified`/`removed` still ignored to match `--diff-filter=A`, and a `docs/.changeset/` lookalike still doesn't match.
- `listFiles` is paginated via `github.paginate`, matching the existing `listComments` call.
- It reports the diff against the **merge base**, which is a more accurate basis for "did this PR add a changeset" than the previous two-dot `git diff`.
- Computing the flag in JS also removes the `${{ steps.detect.outputs.has-changeset }}` expansion that was being interpolated into the script body.
- Net effect: two steps deleted, 8 fewer lines.

### Heads-up on verification

`pull_request_target` always loads the workflow from the base branch, so **this PR will still show the old failure** the fix can only take effect once it's on `main`. Worth a look at the workflow run on the next fork PR after merging to confirm it goes green.

### DIFF
--- a/.github/workflows/changeset-check.yml
+++ b/.github/workflows/changeset-check.yml
@@ -4,6 +4,11 @@ name: Changeset Check
 # changeset, leave a single comment explaining how to add one. If a changeset is
 # later added, the comment is removed. This never fails the build - it only
 # keeps the release-notes pipeline fed.
+#
+# The PR's file list is read through the API rather than checked out. This job
+# runs as `pull_request_target`, so it carries the base repository's token and
+# secrets, and pulling fork code into that trusted context is what leads to
+# "pwn request" vulnerabilities. The check only ever needs filenames.
 
 on:
   pull_request_target:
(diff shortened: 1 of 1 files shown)


addyCooks:
### Expected, this run isn't using this PR's file

`pull_request_target` always loads the workflow from the **base branch**, never the PR head. 
That's the point of the trigger: otherwise anyone could PR a workflow edit and have it run against the repo's secrets.

So what ran here is `main`'s old copy. 
The steps prove it:

```
step 2  Checkout PR head               -> failure
step 3  Detect a changeset in this PR  -> skipped
```

This PR **deletes** both. Its file has one step and no `actions/checkout` anywhere, so it can't produce a failing checkout. The `ref: eeff805…` in the log is just the SHA the old workflow was aimed at.

Net: a `pull_request_target` workflow can't test its own fix. This goes green once it's on `main` the next fork PR is the real check (one step, no checkout, and the "No changeset found" nudge starts working again; it hasn't posted at all since the job started dying at step 2).

Low risk either way: the job is non-blocking by design, so it can't fail a build or affect a release, and reverting one file restores current behaviour.

PR #122, opened 2026-08-26
feat: add streaming proxy mode

### LINKED ISSUE #91: [Feature] Streaming / Pipe-Through Proxy Mode
## Description

Add a `proxy` command that runs a local HTTP proxy to intercept and scrub LLM API requests in-flight.

## Use Case

Instead of manually scrubbing text before pasting it into an API call, developers want a seamless "AI Gateway" that automatically strips PII from outgoing requests and rehydrates incoming responses transparently.

## Proposed Solution

`prompt-scrub proxy --target https://api.openai.com --port 8080`

The proxy would:
1. Intercept outgoing requests to standard LLM endpoints (e.g., OpenAI).
2. Scrub `messages[].content`.
3. Forward to the real API.
4. Intercept the response and rehydrate placeholders before returning it to the user's application.

## Alternatives Considered

Building separate integrations for every LLM SDK, which is unscalable. A proxy works at the network layer and supports all tools.

## Additional Context

akramcodez:
## Description
Closes #91

This PR introduces a local `prompt-scrub proxy` command that acts as a transparent AI Gateway. Instead of manually scrubbing text before calling an LLM API, developers can point their SDKs at this local proxy. It intercepts outgoing requests to standard LLM endpoints (like OpenAI and Anthropic), automatically strips PII from the outgoing payload (`messages[].content`), and flawlessly rehydrates the incoming response before returning it to the client.

Features included:
- `prompt-scrub proxy --target <url> --port <port>` command.
- Transparent intercept & scrub functionality.
- Full streaming (SSE) response rehydration with robust chunk buffering.
- Provider shape detection (OpenAI/Anthropic).
- Integration with the existing session store.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing Notes
- **Automated tests run**: The automated test suite has been expanded, including unit and E2E tests for the proxy mode. All 294 tests across the repository pass perfectly.
- **Manual verification steps**: Manually tested using a custom End-to-End node script that spins up a mock LLM server and issues a client request with PII. Verified that the mock upstream server received the completely scrubbed data (`«Email_1»`) and the client application received the seamlessly rehydrated response (the original email text), including via chunked streaming.

## Checklist


will-lamerton (changes requested):
Nice work - the architecture here is solid. The transform/server split is clean, the transforms are total, provider detection is path-anchored with unknown shapes passing through, and the SSE buffering design (holding back incomplete events, `StringDecoder` for split multibyte) is the right approach.

I checked out the branch and confirmed all 294 tests pass. But I found three correctness bugs that the fake upstream in the tests happens not to exercise.

## Blocking

**1. Stale `content-length` corrupts every non-streaming response** (`src/proxy/proxy-server.ts`, `filterResponseHeaders`)

`content-length` is forwarded verbatim. Rehydration makes the body longer (`«Email_1»` -> `alice@example.com`), so the proxy advertises the upstream's length and then writes more bytes. Reproduced against this branch with an upstream that sets an explicit `Content-Length`, as OpenAI and Anthropic both do:

```
UPSTREAM SAW: {"messages":[{"role":"user","content":"mail «Email_1»"}]}   <- scrub works
Error: Parse Error: Expected HTTP/   (HPE_INVALID_CONSTANT)
```

The client doesn't get a truncated body, it gets a hard HTTP parse failure - the overflow bytes spill past the declared length and desync the parser on the keep-alive connection. Every non-streaming completion containing PII fails.

The tests miss it because `startFakeUpstream` uses `res.end(string)` with no explicit `Content-Length`, so Node falls back to chunked encoding and `transfer-encoding` is stripped as hop-by-hop.

Fix: `delete headers['content-length']` in `filterResponseHeaders`, same as `buildUpstreamHeaders` already does outbound. Please add a regression test with an upstream that sets it explicitly.

**2. `--no-gc` is inert** (`src/cli/commands/proxy.ts`)

Commander maps `--no-gc` to `gc`, never `noGc`. Verified with this repo's commander 15:

```
proxy --no-gc  -> {"gc":false}
proxy          -> {"gc":true}
```

So `if (!options.noGc)` reads `undefined` and GC always runs. Rename the field to `gc?: boolean` and invert the check. The e2e test only asserts the flag shows up in `--help`, so it passes either way.

**3. Streaming is decided from the request, not the response** (`detectStreaming`)

It matches `/"stream"\s*:\s*true/` against the request body. When the upstream answers such a request with a JSON error instead (bad model, rate limit, bad key), the error body goes through the SSE rehydrator:

```
status: 400
body: "{\"error\":{\"message\":\"bad model\",\"param\":\"«Email_1»\"}}\n\n"
placeholder rehydrated? false
```

Spurious `\n\n` appended, and the placeholder leaks unrehydrated. Combined with #1 it also truncates when the error carries a `Content-Length`. The response `content-type` alone is authoritative - dropping the request-body heuristic removes the failure mode and the `requestBody` param.

## Non-blocking

- **Duplicated rehydration.** `response-transform.ts` reimplements `rehydrateString`, which already exists in `src/core/rehydrate.ts`. The core version regex-scans for placeholders actually present; this one loops over every session key and calls `split().join()` unconditionally, allocating even when the token is absent. That's O(session size) scans per SSE chunk on the hot path. It also drops the core version's warnings for hallucinated placeholders. Please reuse `rehydrate()`.

- **Sessions are in-memory only.** Passing `sessionMap` to `scrub()` puts `SessionManager` in stateless mode (`diskEnabled = false`), so nothing persists. So "integration with the existing session store" isn't quite accurate: proxy sessions never show up in `prompt-scrub sessions list`, the startup `gcSessions` call prunes unrelated on-disk sessions and does nothing for the proxy, and the `sessions` Map is never evicted (one PII-bearing entry per header-less request, unbounded). Either persist via `sessionId` or add TTL/LRU eviction, and update the docs either way.

- **Client-supplied session IDs are unvalidated.** Any value is accepted as a map key, and a known session ID lets a caller have arbitrary placeholders rehydrated into real PII. Generated UUIDs are unguessable so this is hardening, not an open hole, but rejecting non-UUIDs closes it cheaply and caps the growth vector above.

- **Target path prefix is dropped.** `dispatchUpstream` sends `path: incomingUrl` and ignores `target.pathname`, so `--target https://gateway.example.com/openai` silently loses `/openai`. Worth joining, since the `detectProvider` comment calls out gateway-prefixed paths as supported.

- **Unconditional SSE re-serialisation.** `JSON.stringify(transformed)` runs even when nothing changed. `rehydrateOpenAIStreamChunk` already returns the original reference in that case, so `if (transformed === parsed) return payload;` skips the work and preserves upstream bytes exactly.

- **No upstream timeout** in `dispatchUpstream` - a hung upstream holds the client socket indefinitely.

- **413 path** responds while the client may still be uploading, which usually surfaces as `ECONNRESET` rather than the intended 413. A `req.destroy()` after responding is more predictable.

## Tooling

- **`knip.json`.** I ran knip both ways. The `"entry": ["src/proxy/index.ts!"]` line exists only to silence 3 unused exports and 6 unused exported types from `src/proxy/index.ts` - nothing consumes the barrel, the CLI imports `runProxy` directly. Suppressing that globally is the wrong lever. Either export the proxy from `src/index.ts` so it's genuinely public API (the docs already reference a programmatic API), or delete the barrel and drop the knip change.

- **`pnpm-workspace.yaml`.** The `auditConfig` change is unrelated to proxy mode - please revert or explain.

## Docs

- `docs/getting-started/cli.md` says "Tune with `--max-body-bytes`", which reads as a CLI flag. It's `maxBodyBytes` on `ProxyOptions` only.
- The garbage-collection and session-continuity sections describe disk-backed behaviour that doesn't match the implementation.

## Tests

Transform coverage is genuinely thorough. Gaps map to the bugs above: no upstream setting `Content-Length`, no behavioural `--no-gc` test, no `stream:true` answered with a JSON error, no concurrent-session test.

Two hygiene points in `tests/proxy/proxy-server.test.ts`: `test.after.always` has an empty body and leaves `.tmp-proxy-config` behind, and assertions inside the fake upstream handler will hang the test rather than fail it, since a throw means the response never gets sent.

---

The scrub direction is correct - I confirmed the upstream only ever sees `«Email_1»`. It's the response direction that's broken against any upstream sending `Content-Length`, which is both providers this targets, and the green suite is masking it. Items 1-3 are small localised fixes; the rest can follow.

will-lamerton:
Hey @akramcodez - just bumping this up for you

Issue #88, opened 2026-08-14
[Feature] `--json` Output Flag for Machine-Readable CLI Output


akramcodez:
## Description

Add a `--json` or `--format json` flag to the CLI commands (`scrub`, `inspect`, `rehydrate`) to output results as a structured JSON object.

## Use Case

Currently, the CLI outputs human-readable text. Developers building CI/CD pipelines, automated scripts, or programmatic wrappers (like AI Gateways) need a machine-readable format to easily parse the scrubbed output, session map, and statistics without brittle string parsing.

## Proposed Solution

When the `--json` flag is provided, output a JSON object containing the `scrubbedContent`, `sessionId`, `sessionMap`, and basic stats.

## Alternatives Considered

Relying on standard stdout/stderr separation (which is currently used), but this doesn't solve the need for structured metadata mapping alongside scrubbed output. Piping stdout and parsing it with `sed`/`grep` is extremely brittle and breaks when the text itself contains special characters.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first, privacy-respecting tooling)

sumitM-263:
@akramcodez I'd like to work on this issue. Can I be assigned?

akramcodez:
sure @sumitM-263 go for it :)

PR #125, opened 2026-08-27
feat: add JSON output to CLI commands

sumitM-263:
## Description
Adds a `--json` option to the `scrub`, `inspect`, and `rehydrate` CLI commands. This provides machine-readable output for CI pipelines, scripts, and programmatic integrations.

- The JSON output avoids exposing the sensitive `sessionMap` by default. `scrub --json --include-session-map` explicitly enables it (documented as writing plaintext PII to stdout/CI logs, and errors out if used without `--json`).
- All three commands emit their primary payload under a unified `content` key.
- `sessionId` is present only when something was scrubbed and a session was persisted — omitted (not empty) otherwise.
- Success JSON goes to stdout; the `{ "error": ... }` envelope (stdin/file read failures) goes to stderr with exit code 1, so `... --json | jq` is safe.
- `inspect --json` composes with `--min-confidence`: findings dropped by the threshold are listed in a `suppressed` array (value, span, confidence, method) rather than silently vanishing — the same guarantee the text output gives, so a filtered JSON report can't read as "safe" while dropped values remain in the clear.
- Also fixes inspect's placeholder numbering to match actual scrub output (right-to-left replacement with per-value deduplication); inspect text and JSON output now show the same placeholders scrub will produce.

Closes issue: #88

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Refactoring

## Testing Notes
- Added CLI end-to-end tests for JSON output.
- `pnpm run test:types` passes.
- `pnpm run test:lint` passes with an existing Biome schema-version informational message.
- `pnpm ava tests/cli/e2e.test.ts` passes: **59 tests** (grew from 42 after the merge brought in the `--min-confidence` e2e tests, plus a new combined `--json --min-confidence` test).
- `pnpm ava tests/cli/inspect.test.ts`: unit tests pass (13 tests, including the `--min-confidence` unit tests merged from upstream); the "fails when no stdin is provided" test times out on Windows only (`readFileSync(0)` blocks on the ava worker's stdin). Pre-existing and unrelated to this PR — passes on Linux CI.
- JSON test coverage: repeated-value placeholder deduplication; inspect `--json` placeholders matching actual scrub output on a 3-entity input; inspect `--json` with `--min-confidence` reporting suppressed findings; scrub `--json` → rehydrate `--json` round-trip using the returned `sessionId`; `sessionId` omitted when nothing is scrubbed; empty input; file input; error envelopes, exit codes, and stderr routing; `--include-session-map` gating.
- Local `biome ci` on Windows reports line-ending-only differences (CRLF working tree from `core.autocrlf`); committed content is LF and repository format checks pass in CI.
- Repository-wide `pnpm run test:format` still reports existing formatting differences in unrelated files; no unrelated files were reformatted.

## Checklist
- [ ] I have formatted, linted, and passed all tests (`pnpm run check`).


will-lamerton (changes requested):
Thanks for this. I built the branch and exercised the flag against real input. Build, `test:lint`, `test:format`, and all 26 e2e tests pass. Two things need resolving before merge.

## 1. `scrub --json` prints every original value to stdout

```
$ printf 'Email alice@example.com and bob@example.com' | prompt-scrub scrub --json
{
  "scrubbedContent": "Email «Email_2» and «Email_1»",
  "sessionMap": { "«Email_1»": "bob@example.com", "«Email_2»": "alice@example.com" },
  ...
}
```

The text path deliberately keeps stdout clean so it is safe to pipe (content on stdout, session ID and summary on stderr). `docs/getting-started/cli.md` states that contract. `--json` inverts it in exactly the context this feature targets: `prompt-scrub scrub --json | llm` now transmits the raw values.

I know #88 asked for `sessionMap` in the payload, so this is the spec rather than a slip, but the map is already persisted and retrievable via `sessions show <id>`, so inlining it is a convenience rather than a capability. Suggest omitting it by default and gating it behind an explicit opt-in such as `--include-session-map`, with the hazard documented. Happy to defer if a maintainer wants it on by default, but it should be a deliberate call.

## 2. `inspect --json` placeholders do not match what `scrub` produces

`toInspectJson` numbers findings left-to-right, one per finding. Real scrubbing runs right-to-left and dedupes by value (`SessionManager.createPlaceholder` reuses the placeholder for a repeated value). Both axes diverge:

```
input: "Email alice@example.com and alice@example.com and bob@example.com"

inspect --json:  Email_1 (alice), Email_2 (alice), Email_3 (bob)
scrub  --json:   Email_2 (alice), Email_2 (alice), Email_1 (bob)
```

Every placeholder there is wrong. It is inherited from `formatInspectOutput`, where it is cosmetic, but as a machine-readable contract it becomes a correctness bug: anything mapping `entities[].placeholder` onto the scrubbed text will mismatch.

The fix is already in the same file. `computeHash` does the correct simulation with a throwaway `SessionManager`, and I confirmed its hash matches real `scrub` output. Reuse that loop to capture the finding-to-placeholder mapping and have both `toInspectJson` and `formatInspectOutput` read from it. That fixes the text output too and drops the duplicated counter.

## Smaller items

- **Docs.** `docs/getting-started/cli.md` lists options per command; `--json` is missing from all three, and the output shapes (`entities`, `stats`, the `{error}` envelope) are undocumented. For a flag meant to be a stable contract for scripts, the schema needs writing down.
- **Empty input emits no JSON.** `if (!input) process.exit(0)` short-circuits before any output, so `printf '' | prompt-scrub scrub --json` exits 0 with empty stdout and the consumer's `JSON.parse` throws. JSON mode should emit a valid empty result.
- **Errors go to stdout in JSON mode.** Defensible, but it means callers must parse stdout to tell success from failure. Exit code is already 1, so I would keep the error JSON on stderr. Either way, document the choice.
- **Six copies of the same error block.** The `if (options.json) { ... } else { console.error }` block is pasted twice per file across three files. A small `src/cli/output.ts` with an `emitError(message, json)` helper would collapse all six and keep the envelope consistent.
- `toInspectJson` has no explicit return type, unlike `computeHash(): string` and `formatInspectOutput(): string` next to it. An exported interface would match local style and give the contract a name to document.
- The `// Read from stdin` comment removals are unrelated to the feature, and `scrub.ts` keeps its copy, so the three files are now inconsistent.

## Tests

Three happy-path tests, all green. The gaps line up with the bugs: no repeated-value case (would have caught #2), no coverage of the `{error}` envelope or its exit code, no empty-input case, and no `--json` with a file argument rather than stdin.

## What is good

`rehydrate --json` is a real improvement over the text path, which strands warnings on stderr for callers to correlate by hand. Changeset is correct at `minor`, `package.json` version untouched as the contributor guide asks, and the flag definitions match the surrounding commander style.

sumitM-263:
Addressed the requested changes:

- Omit `sessionMap` from `scrub --json` by default.
- Add `--include-session-map` as an explicit opt-in.
- Align `inspect --json` placeholders with actual scrub behavior.
- Handle empty JSON input consistently.
- Send JSON errors to stderr with exit code 1.
- Centralize error output through `src/cli/output.ts`.
- Added documentation and expanded CLI tests.
- Verified 38 e2e tests pass, along with type checking, linting, and formatting checks for the changed files.

will-lamerton (changes requested):
Nice feature, and the default of withholding `sessionMap` is the right call. I built the branch and ran `biome ci` (clean) and `ava tests/cli` (45 pass), then exercised the JSON paths by hand. A few things to fix first.

### Blocking

1. **`scrub --json` emits a `sessionId` for sessions that were never persisted.** `scrub()` only calls `session.save()` when the map is non-empty, and the text path mirrors that by printing `Session ID:` only when the content changed. The JSON path emits it unconditionally:
   ```
   $ printf 'hello world' | prompt-scrub scrub --json
   { "scrubbedContent": "hello world", "sessionId": "fb21a1fa-...", "stats": {...} }
   ```
   That UUID is not on disk, so a script that stores it gets an empty map back from `rehydrate`. Please omit `sessionId` when nothing was scrubbed.

2. **Empty input emits `"sessionId": ""`**, which is not a usable ID. Same fix.

3. **Docs overstate the error envelope.** Only the file/stdin read sites call `emitError`; anything thrown later hits the top-level `catch` in `src/cli/index.ts` and dumps a raw `Error`. Either scope the doc to input errors or wrap the action bodies. The docs also do not mention that the envelope goes to **stderr** while success JSON goes to stdout, which matters for `... --json | jq`.

4. **Inconsistent key names across commands:** `scrub` returns `scrubbedContent`, `rehydrate` returns `content`; `rehydrate` has `warnings`, the others do not. This is a public contract, worth settling now.

### Worth calling out

The `formatInspectOutput` change is a real bug fix that is not mentioned anywhere. `scrubString` replaces right to left, so `scrub` produces `Email_3, Email_2, Email_1` for three emails, while the old `inspect` counters showed `Email_1, Email_2, Email_3`. Reusing the `SessionManager` walk makes them agree. Please add a test pinning that (inspect placeholders match scrub output for a 3-entity input) and mention the fix in the changeset.

### Smaller items

- `computeHash('', [])` instead of hardcoding `e3b0c442...b855` in both `inspect.ts` and the test.
- `output.ts` has only `emitError`, yet `process.stdout.write(JSON.stringify(x, null, 2))` is copy-pasted six times. Add `emitJson()`.
- `placeholderMap.get(i) || '«UNKNOWN»'` is unreachable, and the `if (!finding) continue` guards only exist to satisfy `noUncheckedIndexedAccess`. Returning `Array<{finding, placeholder}>` from `computeHash` lets both call sites keep the original `for...of` and drops the guards and the sentinel.
- **Please revert the unrelated churn in `inspect.ts`.** Lots of added blank lines between adjacent statements, and several useful comments deleted (`// Read from stdin` x3, `// Use a dummy session to simulate exactly what scrub does`, `// Truncate very long values for display`). It roughly doubles the size of the diff.
- `--include-session-map` is silently ignored without `--json`. Note "requires `--json`" in the option description, or error.
- Docs: stray space in `` `--json` : `` (elsewhere it is `` `--hash`: ``), trailing whitespace, no newline at EOF, `//` comments inside a JSON code block, and the `⚠️` emoji is not used elsewhere in these docs. Worth adding that `--include-session-map` puts plaintext PII into CI logs.
- Tests: leftover `// ← CHANGED from t.notEqual`; file fixtures are written into `tmpConfigDir` rather than a temp dir; no round-trip test (scrub `--json` then rehydrate with the returned `sessionId`), which would have caught item 1.

No security concerns. I confirmed no core code path writes to stdout, so the JSON cannot be corrupted by incidental logging.

akramcodez:
@sumitM-263 Please resolve the merge conflicts and address the review comments, then ping me once it’s ready for another look.

sumitM-263:
Hi @akramcodez,

I've pushed a merge commit (`b56ddcb`) resolving the conflicts with `main` and addressing all the feedback from both review rounds:

* `sessionMap` stays gated behind `--include-session-map` and errors without `--json`; the plaintext-PII/CI-logs hazard is documented.
* Inspect's placeholder numbering now mirrors actual scrubbing (right-to-left replacement, per-value dedup). The fix is mentioned in the changeset and pinned by a test asserting that inspect's placeholders match actual scrub output.
* `sessionId` is omitted when nothing was scrubbed; key names are now consistent (`content` everywhere, with `warnings` only on `rehydrate`).
* The error envelope goes to stderr with exit code 1, is scoped to input errors, and is documented. Success JSON remains on stdout.
* Empty input now emits valid JSON, and a scrub → rehydrate round-trip test has been added.
* The JSON path now also composes with the new `--min-confidence` flag: `inspect --json --min-confidence` includes a `suppressed` array so filtered reports don't appear "safe" while dropped findings remain visible.

The branch is ready for another review. Thanks!

akramcodez:
Congratulations on your first PR merge in Nano Collective, @sumitM-263! Feel free to add yourself as a contributor to our website via a PR, and I’ll approve it.

https://nanocollective.org/contributors
https://github.com/Nano-Collective/organisation

PR #128, opened 2026-08-28
fix(watch): use one session per run so placeholders never collide

### LINKED ISSUE #124: [Bug] watch --file mints a new session every tick, collapsing different values into the same placeholder
## Description

`watchFileStep` forwards its options to `handleScrub` without a session ID unless the user passed
`--session-id`. With none, `SessionManager` generates a fresh UUID and starts from an empty map, so
the per-category counter resets to `1` on **every tick**.

The file on disk still contains `«Email_1»` from the previous tick, and nothing in the pipeline
reserves or even notices an existing placeholder — so the next tick issues that same name for a
different value and writes it back over the file in place.

The result is **unrecoverable data loss in the default documented flow**: once the file reads
`«Email_1»` twice, the distinction between the two values is gone from the only copy that exists.
No session can restore it, because no session ever recorded that the two tokens meant different
things. Every rehydration is confidently wrong rather than obviously broken.

## Environment

- **OS**: Windows 11 (26100) not platform-specific
- **Node version**: v22.22.3
- **Install type**: both (built `dist/` CLI and library API)

## Steps to Reproduce

1. `echo "me: alice@a.com" > notes.txt`
2. `prompt-scrub watch --file notes.txt` (no `--session-id`)
3. Wait for the first tick. `notes.txt` becomes `me: «Email_1»`.
4. Append a second, different address: `echo "them: bob@b.com" >> notes.txt`
5. Wait for the next tick, then inspect `notes.txt` and the sessions directory.

## Expected Behavior

notes.txt   me: «Email_1»
them: «Email_2»

one session, rehydrating to:
me: alice@a.com
them: bob@b.com

## Actual Behavior

notes.txt   me: «Email_1»
them: «Email_1»        ← two different people, one placeholder

two sessions on disk:
1d727b8c….json  { "«Email_1»": "bob@b.com" }
dd0155a1….json  { "«Email_1»": "alice@a.com" }

rehydrate --session-id 1d727b8c…  →  "me: bob@b.com\nthem: bob@b.com"
rehydrate --session-id dd0155a1…  →  "me: alice@a.com\nthem: alice@a.com"

Neither rehydration is correct, and the original distinction is unrecoverable.

## Logs/Screenshots

Two further consequences of the same root cause:

- **A session file accumulates per tick.** Four edits produced four session files; a long-running
  watcher grows the session store without bound.
- **The session ID is never printed.** `watchFileStep` logs only
  `[watch] Scrubbed 1 email in <path>.`, so even in the non-colliding case the user has no way to
  know which of N UUID files rehydrates their file.

Confirmed that passing `--session-id` avoids it the same sequence then produces
`me: «Email_1»\nthem: «Email_2»` and round-trips exactly. So the fix is to allocate one session ID
in `handleWatch` before the first tick (reuse `options.sessionId` when given, otherwise generate a
UUID once), thread it through every `watchFileStep` / `watchClipboardStep` call, and print it on
start-up so rehydration is possible.

Relevant code:

- `src/cli/commands/watch.ts:236` `await handleScrub(current, options)`, no session ID threaded
- `src/session/session-manager.ts:22-26` fresh UUID + empty map when no session ID is supplied

Worth noting the underlying hazard is in the core, not just in watch: `createPlaceholder`
(`src/session/session-manager.ts:89-103`) issues names from a counter with no awareness of the text
being scrubbed. The same collision is reachable in a single `scrub` call on input that already
contains a placeholder e.g. re-scrubbing scrubbed output turns
`"the token «Email_1» is literal, contact a@b.com"` into
`"the token «Email_1» is literal, contact «Email_1»"`, which rehydrates to a real address in both
positions. Fixing watch resolves the reported symptom; having `createPlaceholder` skip any candidate
name already present in the text resolves the class.

## Additional Context

addyCooks:
Closes #124

## Description

`watch` could destroy data by handing one placeholder to two different values.

Because every tick started a fresh session, `«Email_1»` was reissued for a second address and written back over the file the only copy of the original. Once the file read `«Email_1»` twice, no session could tell the two values apart; every rehydration was confidently wrong rather than obviously broken.

**`handleWatch` now allocates one session for the whole run** (`src/cli/commands/watch.ts`) reusing `--session-id` when given, otherwise generating a UUID once threads it through every `watchClipboardStep` / `watchFileStep` call, and prints `[watch] Session ID: <id>` on start-up so rehydration is possible. It prints *after* the clipboard preflight, so a run that dies on a missing `xclip` doesn't announce a session it never used.

**The underlying hazard is fixed in the core, not just in `watch`.** As noted on the issue, `createPlaceholder` issued names from a counter with no awareness of the text being scrubbed, so the same collision was reachable in a single `scrub` call on input that already contained a placeholder. `SessionManager` now records placeholder-shaped tokens present in the text (`reservePlaceholdersIn`, called from `scrubString`) and skips any candidate name that is taken:

the token «Email_1» is literal, contact a@b.com
→ the token «Email_1» is literal, contact «Email_2»
→ rehydrates back to the original, literal token intact

**One knock-on.** With a session that now spans the whole run, `formatNotificationMessage(result.sessionMap)` would have reported the running total on every tick ("Scrubbed 2 emails" when it scrubbed one). It is now driven by `result.stats.byCategory` that scrub's own stats. The function is module-internal (not exported from `src/index.ts`), and its output is unchanged for every previously-tested input.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing Notes

**Automated**  4 tests added, each confirmed to **fail** with the source change reverted and pass with it:

| Test | File |
| --- | --- |
| `handleWatch shares one session so distinct values never collide` | `tests/cli/watch.test.ts` |
| `handleWatch reuses an explicit --session-id instead of minting one` | `tests/cli/watch.test.ts` |
| `a later tick reports only what it scrubbed, not the running session total` | `tests/cli/watch.test.ts` |
| `a placeholder already present in the text is never reissued to a new value` | `tests/core/scrub.test.ts` |

The first drives two distinct emails through one pass, asserts `«Email_1»`/`«Email_2»`, then round-trips both through the *printed* session ID so it fails if the ID stops being printed, too.

**Manual** reproduced the issue's exact steps against the built CLI, with a real long-running watcher (`--interval 300`) and a second address appended between ticks:

| | before | after |
| --- | --- | --- |
| `notes.txt` | `me: «Email_1»` / `them: «Email_1»` | `me: «Email_1»` / `them: «Email_2»` |
| session files | one per tick | one for the run |
| session ID | never printed | `[watch] Session ID: <id>` |
| `rehydrate` | wrong in both positions | `me: alice@a.com` / `them: bob@b.com` |
| tick 2 log | | `Scrubbed 1 email` (not the running total) |

**`pnpm run check`** run in a clean clone of this branch on a space-free path with LF endings. `test:format`, `test:types`, `test:lint`, `test:knip` and `test:audit` all pass; `test:security` is skipped by `scripts/test.sh` (no semgrep locally).

`test:ava` reports 257 passing, 2 failing, 17 pending on my Windows machine. All of that is pre-existing and environmental pristine `main` in the same clone gives **253 passing and the same 2 failures**:

- `deleteSessionMap handles unlinkSync error` makes the dir read-only via `chmodSync(dir, 0o555)`, which does not remove delete rights on NTFS, so the unlink succeeds
- `loadConfiguredRulePacks loads a valid mock pack` dynamic-imports a bare `C:\...` path, which is not a valid ESM specifier (needs `pathToFileURL`)
- the 17 pending are the 3 `*no stdin is provided*` tests, which rely on "the test environment having closed stdin"; ava's Windows worker gets an open socket on fd 0, so `readFileSync(0)` blocks. Excluding them: **259 tests, 257 pass, 2 fail.**

Happy to open separate issues for those three if they'd be useful they are all Windows-only and none are touched by this PR.

## Checklist
- [x] I have formatted, linted, and passed all tests (`pnpm run check`). *(see Testing Notes: the only ava failures are pre-existing Windows-only ones, identical on `main`)*


will-lamerton (changes requested):
Great diagnosis, and the `watch` half of this is exactly right - one session per run, threaded through `stepOptions`, ID printed after the preflight. Full suite passes for me (262/262, macOS / Node 22). Requesting changes only on the core guard, which doesn't quite close the hazard it's aimed at. Both items below I reproduced against this branch's build.

### 1. The reservation regex doesn't match the placeholders we actually mint

`PLACEHOLDER_TOKEN_REGEX = /«[A-Za-z]+_\d+»/g`, but `placeholderPrefix` is a free-form string on the public extension surface (`options.customDetectors`, and `docs/features/authoring-rule-packs.md`). A rule pack using prefix `Ticket2` still collides:

```
in:  literal «Ticket2_1» plus TKT-99
out: literal «Ticket2_1» plus «Ticket2_1»
```

Both tokens rehydrate to `TKT-99` - the exact failure this PR is fixing. Pre-existing (`rebuildCategoryCounts` has the same `[A-Za-z]+` restriction), but a guard whose whole job is closing that hole shouldn't inherit it. Suggest `/«[^«»]+_\d+»/g`, and aligning `rebuildCategoryCounts` in the same pass.

### 2. Reservation is forward-only, so `Message[]` input still collides

`reservePlaceholdersIn` runs per `scrubString`, so it only protects against literals seen at or before the current message. Reverse the order of your own test case and it breaks:

```js
scrub({ content: [
  { role: 'user', content: 'real x@y.com' },
  { role: 'user', content: 'literal «Email_1» here' },
]})
// -> ["real «Email_1»", "literal «Email_1» here"]  two meanings, one token
```

Cheap to close properly: hoist the reservation into `scrub()` and reserve over the whole request before scrubbing any part of it.

```ts
// in scrub(), before the string / Message[] branch
if (typeof content === 'string') session.reservePlaceholdersIn(content);
else for (const msg of content) session.reservePlaceholdersIn(msg.content);
```

That subsumes the current `scrubString` call site and makes the guard order-independent. Worth a test on the reversed-message case, plus one for the non-alphabetic prefix above.

### 3. Non-blocking: the stats swap changes counts for repeated values

`stats.byCategory` counts findings; the session map deduplicates by value. So `'a@b.com and again a@b.com'` goes from `Scrubbed 1 email` to `Scrubbed 2 emails`. The new number is arguably the better one, but "output is unchanged for every previously-tested input" holds only for the tests - worth a line in the changeset. Category naming itself is safe: `category === placeholderPrefix` for all eight built-in detectors.

### Nits

- `--dry-run` watch still persists a session file (`scrub` saves whenever the map is non-empty) and now announces its ID on start-up. Pre-existing and out of scope, just a bit odd to advertise a session a dry run arguably shouldn't have written.
- Tests are strong - round-tripping through the *printed* session ID pins that log line as contract rather than incidental output, and the `logCount` to `logs[]` refactor is a real improvement.

addyCooks:
Thanks @will-lamerton,
both blocking items are fixed, 
each reproduced against the previous build first.

**1. Prefix regex.** Widened to `[^«»]+`. It was in **three** places, not two: `rehydrate`'s own `PLACEHOLDER_REGEX` had the same restriction, so a `Ticket2` placeholder was scrubbed away and then *never restored*. Fixing only the two you named would have turned silent corruption into a token that round-trips to nothing. All three now match, and I documented the one real constraint (no `«`/`»`) for pack authors. Also checked the wider class for backtracking linear, 800 KB in 6 ms, so no ReDoS concern.

**2. Forward-only reservation.** Hoisted into `scrub()` as you sketched. Your reversed-message case now gives `["real «Email_2»", "literal «Email_1» here"]` and round-trips.

**3. Stats swap.** Confirmed 1 → 2 for a repeated value. Now in the changeset as a behaviour change, with a test pinning it.

**`--dry-run`:** left alone per "out of scope". The two halves point opposite ways the session file is written either way, so printing the ID makes an already-written session identifiable; suppressing just the announcement would leave an unidentifiable file. The real fix (dry runs not persisting) needs `handleScrub` plumbing  happy to do it as a follow-up.

**Tests**, four added, each verified to fail with its fix reverted. My first counter test passed even with the bug, because the `in this.map` fallback caught it; rewrote it as a gap case (`«Ticket2_5»` → `«Ticket2_6»`) so it actually discriminates. Also re-ran the issue's repro end-to-end: `«Email_1»`/`«Email_2»`, one session file, ID printed, both originals restored.

**Semgrep isn't this PR** `main` has 21 findings and the job runs `--error`, so every PR shows it (non-blocking; run concludes `success`). Diffed against main: zero new. Everything else green, including Unit Tests and knip.

Ready for another look.

will-lamerton (changes requested):
Diagnosis is right and the fix is in the right layer - fixing only `watch` would have left the same hazard reachable from the library. Reserving across the whole request rather than per message is the correct call, and widening `rebuildCategoryCounts` and `rehydrate` alongside the guard is the non-obvious part a narrower fix would have missed. Verified at `aa3c589`: 266 ava tests pass, `tsc --noEmit` clean.

A few things to fix before merge.

### 1. `formatNotificationMessage` re-implements pluralization badly, next to a correct one

`src/cli/commands/scrub.ts` already has `pluralize()` and `formatScrubSummary` builds this exact breakdown from the exact same `stats.byCategory`. This PR moves `formatNotificationMessage` onto that same input but keeps a naive `${name}s`:

```
formatNotificationMessage({ Address: 2, Secret: 2, Url: 2 })
-> "Scrubbed 2 addresss, 2 secrets, 2 urls"
```

`PostalAddressDetector` uses `category: 'Address'`, so `watch` prints and notifies "Scrubbed 2 addresss" in a real run. Export `pluralize` and call it:

```ts
const parts = entries.map(([category, count]) => `${count} ${pluralize(category.toLowerCase(), count)}`);
```

The test at `tests/cli/watch.test.ts:36` only covers `Secret`/`Email`, the two categories where naive `+s` happens to be right. Please add `Address`.

### 2. Widened rehydrate regex warns on ordinary guillemet text

`[^«»]+` matches whitespace, so any French or Russian quotation ending in `_<digits>` is now read as a placeholder:

```
rehydrate({ content: 'il dit «bonjour mon ami_1» ok', sessionMap: {} })
-> warnings: ['Warning: placeholder «bonjour mon ami_1» not found in session - left as-is.']
```

Content is unchanged so this is noise, not corruption, but prompt content is arbitrary user text. `[^«»\s]+` still admits `Ticket2` and `Jira_Issue`. Please apply to all three patterns and note the no-whitespace constraint in the `authoring-rule-packs.md` paragraph you added.

### 3. Changeset overclaims, and will render as broken markdown

`createPlaceholder` returns `getExistingPlaceholder(originalValue)` before the reserved-name loop, so a literal token that is also a map key still collapses:

```ts
const m = { '«Email_1»': 'alice@a.com' };
scrub({ content: 'literal «Email_1» and alice@a.com', sessionMap: m });
// -> 'literal «Email_1» and «Email_1»', rehydrates to 'literal alice@a.com and alice@a.com'
```

Narrow in practice (both positions rehydrate to the value the token already denoted), so I am not asking for a code change - just soften "re-scrubbing scrubbed output no longer collapses a literal «Email_1» and a real address into one token", which is not quite true.

Separately, `.changeset/changelog.cjs` prefixes `- ` to a non-list summary. At four paragraphs, only the first lands inside the bullet and the rest de-indent in `CHANGELOG.md`. Every other changeset here is one or two sentences - please trim to roughly the first paragraph plus the behaviour-change note.

### Nits

- `newPlaceholder in this.map` walks the prototype chain on a caller-supplied object. No practical impact, but `Object.hasOwn` states the intent.
- `options.sessionId ?? randomUUID()` accepts `--session-id ""`, giving a session path of `sessions/.json`. Pre-existing, but this PR makes the ID load-bearing for the whole run.
- `watch --dry-run` still persists a session map even though nothing is written. Pre-existing, now more visible under an announced ID. Follow-up, not this PR.

No security or performance concerns: one linear regex pass per request, patterns anchor on a literal `«` and exclude both guillemets so there is no backtracking blowup, and the counter advances past skipped names so the skip loop stays amortized linear.

addyCooks:
Thanks @will-lamerton, 
all three fixed in e9ad45d:

1. **Pluralization.** `formatNotificationMessage` now calls the real `pluralize()` from `scrub.ts`. Added the `Address` case to the test.
2. **Whitespace in the prefix regex.** Widened to `[^«»\s]+` in all three places (reservation, counter, rehydrate) and noted the no-whitespace constraint in the rule-pack doc. Added a test pinning that ordinary guillemet-quoted prose no longer warns.
3. **Changeset.** Trimmed to one paragraph, dropped the "no longer collapses" overclaim.

Nits taken too: `Object.hasOwn` instead of `in`, and `--session-id ""` now falls back to a generated ID instead of writing to `sessions/.json`. Left `--dry-run` persistence alone as you suggested.

Also merged `upstream/main` in the branch was 15 commits behind and dragging in stale Semgrep findings unrelated to this PR (fixed in 737a055).

Ready for another look.

Issue #92, opened 2026-08-14
[Feature] prompt-scrub diff - Visual Side-by-Side Comparison


akramcodez:
## Description

Add a `diff` command that outputs a visual, colorized diff of the original text vs. the scrubbed text.

## Use Case

The `inspect` command shows a table of detected entities, but users building trust with a new privacy tool want to see exactly what is being changed in context — similar to `git diff`. This is more intuitive than a table: it shows the before/after inline, making it immediately obvious if the tool is changing something it shouldn't. This builds confidence and is the ideal "first command" for new users.

## Proposed Solution

Add a `diff` subcommand. Usage example:

    echo "Email me at alice@corp.com" | prompt-scrub diff

Proposed output (with terminal colors — red for removed PII, green for placeholder):

    - Email me at alice@corp.com
    + Email me at «Email_1»

Support additional flags:
- `--side-by-side` for a two-column layout
- `--context <n>` to control surrounding context lines (default: 3)
- `--no-color` for piping to files or non-TTY contexts

## Alternatives Considered

Relying purely on `inspect`, which lists findings in a table but doesn't show the structural impact on the prompt in-line. Users still have to mentally reconstruct what changed. Using `diff <(cat original) <(prompt-scrub scrub original)` works but requires two commands and loses the coloring.

## Additional Context

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] This feature aligns with the project's goals (local-first, privacy-respecting tooling)

addyCooks:
Hey @akramcodez!  
can I get this assigned?

What I'm thinking to do:
- No diff lib `Finding` already has exact char spans, so before/after is derived, not inferred. 
Zero new deps.
- Reuse `inspect`'s simulate-without-save path so `diff` stays a real dry run.
- Tiny ANSI helper, auto-off on non-TTY + `NO_COLOR`.
- `setupDiffCommand` in `src/cli/index.ts` mirroring `watch`, tests in `tests/cli/`.

Also noticed: `scrub` numbers placeholders in reverse (`«Email_2»` before `«Email_1»`)  harmless in the table, obvious in a diff. 
Can file separately if you'd rather keep this PR additive.

Dhirenderchoudhary:
@addyCooks can i work on this

akramcodez:
Sure, @Dhirenderchoudhary, go for it! I just assigned @addyCooks to another issue.

PR #129, opened 2026-08-28
Add a diff command for original vs scrubbed text.

Dhirenderchoudhary:
## Description
Adds `prompt-scrub diff` for issue #92. `inspect` already lists findings in a table; this prints a git-style view of original vs scrubbed text so you can see replacements in context. Dry run only: no session is written.

```bash
echo "Email me at alice@corp.com" | prompt-scrub diff --no-color
```

```
- Email me at alice@corp.com
+ Email me at «Email_1»
```

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing Notes
- Automated: `pnpm run test:format`, `test:types`, `test:lint`, `test:ava` (269 tests), `test:knip`, `test:audit`. Unit tests in `tests/cli/diff.test.ts` plus e2e coverage for the default output, `--side-by-side`, no session write, and `--help`.
- `pnpm run check` / `test:security` still fails on existing GitHub Actions/semgrep findings in this repo (mutable action tags, etc.). Not introduced here.
- Manual: `echo "Email me at alice@corp.com" | prompt-scrub diff --no-color` and the same with `--side-by-side` / `--context 1`.

## Checklist


will-lamerton (changes requested):
Nice addition - it slots cleanly next to `inspect`, writes no session, and the `simulateScrub` extraction is behaviour-preserving (verified it matches real `scrub()` output). Two things to fix before merge.

### 1. `--side-by-side` truncation can hide unscrubbed PII

`cell()` hard-truncates each line to the column width:

```js
formatDiff('prefix '.repeat(20) + 'unscrubbed-secret@corp.com', ..., { sideBySide: true, width: 60 })
// "prefix prefix prefix pref... | PFX prefix prefix prefix ...\n"
```

The email is gone from the output entirely. For a tool whose job is "look at this before you send it", a view that silently drops the tail of a line is a correctness problem - a user can conclude nothing was missed when something was. Wrap instead of truncating, or at minimum emit a visible marker plus a stderr note that N lines were truncated.

### 2. `diffLines` is O(n x m) in time and memory

The full LCS table is materialised as `(n+1) x (m+1)` numbers. Measured on this branch:

| lines | time | RSS |
|---|---|---|
| 4,000 | 197 ms | 139 MB |
| 8,000 | 777 ms | 632 MB |
| 12,000 | 2.1 s | 1.2 GB |

~30k lines OOMs, and pasting a large log or file dump into a scrubber is a normal use case. Cheapest fixes: trim the common prefix/suffix of identical lines first (scrubbing is in-place, so usually only a few lines differ), and/or fall back to index-wise pairing when `a.length === b.length`.

### Non-blocking

- **ANSI passthrough**: input lines are written verbatim, so `\x1b[2J\x1b[H` in the input clears the screen and hides the content being inspected; an embedded `\x1b[0m` also ends the colour run early. Suggest stripping C0 controls and `\x1b` (keeping `\t`) before rendering - `inspect` has a milder version of the same issue, so a shared sanitiser would cover both.
- **`inspect` vs `diff` disagree on placeholder numbers** (pre-existing, but this PR makes it visible and the docs now chain the two commands). For `Mail alice@corp.com then bob@corp.com`, `inspect` says alice -> `«Email_1»`, but `diff` (and real `scrub`) emit `Mail «Email_2» then «Email_1»`. `diff` is the correct one; `formatInspectOutput` numbers left-to-right while `scrub` creates placeholders during its right-to-left pass. Worth a follow-up issue.
- **Docs**: `docs/getting-started/cli.md` documents 4 of the 8 options - `--enable`, `--strict-name`, `--code-tell-terms`, `--url-allowlist` and `NO_COLOR` are implemented but undocumented. The same file still describes `inspect` as printing "a human-readable diff", which is now confusing. The README edit also dropped the sentence about the hash being deterministic for cache-stability checks - unrelated to this change, worth keeping.
- **Duplication**: the stdin/file-read block and the five detector options are now triplicated across `scrub.ts`, `inspect.ts` and `diff.ts`. Good point to extract `readInput(file)` and an `addDetectorOptions(cmd)` helper.
- **`parseContext`** silently coerces `--context abc` and `--context -5` to `3`; commander's `InvalidArgumentError` would be more honest.
- **Tests**: the help test (`stdout.includes('diff')`) would pass on the word appearing in any description - assert the description string instead. Gaps worth covering: invalid `--context`, inputs where line counts differ (a finding value spanning a newline), and the `...` separator between non-adjacent hunks.

will-lamerton (changes requested):
Both blockers are properly fixed - verified on `fb9a99c`: side-by-side now wraps and keeps the tail email intact, and 30k lines diffs in 5 ms / 68 MB (was 1.2 GB at 12k). Docs, `parseContext`, and the test gaps are all addressed. 277 tests, lint and knip clean.

One new issue to fix before merge.

### `sanitizeLine` swallows input after a CSI sequence with a non-alphabetic final byte

`src/cli/sanitize.ts:9` scans with `!/[A-Za-z@]/`, but CSI final bytes are the whole `0x40-0x7E` range, which includes `~ { | } ` and `<=>?`. When the final byte is one of those, the scan runs past it and eats real content until it hits a letter, or to end of line if there is none:

```js
sanitizeLine('\x1b[3~leaked-secret@corp.com')  // "eaked-secret@corp.com"  (the "l" is gone)
sanitizeLine('\x1b[1}visible@corp.com')        // "isible@corp.com"
sanitizeLine('\x1b[3~+1 415-555-1234')         // ""   <- entire phone number vanishes
```

`\x1b[3~` is just the Delete key, so it shows up in pasted terminal captures routinely. This is the same failure mode as the original side-by-side blocker: the line renders clean and the user concludes nothing was missed. It affects `inspect` too now that both share the sanitiser.

Minimal fix keeping the loop shape:

```ts
if (code === 27) {
  if (text[i + 1] === '[') {
    i += 2;
    // CSI: parameter bytes 0x30-0x3F, intermediates 0x20-0x2F, terminated by 0x40-0x7E
    while (i < text.length) {
      const c = text.charCodeAt(i);
      if (c >= 0x40 && c <= 0x7e) break;          // final byte; the for-loop's i++ consumes it
      if (c < 0x20 || c > 0x3f) { i--; break; }   // malformed, do not eat real content
      i++;
    }
  }
  continue;
}
```

`sanitize.ts` is a new shared module with no test file of its own - the single assertion in `diff.test.ts` only covers `\x1b[2J`. Worth a `tests/cli/sanitize.test.ts` covering the final-byte range, a truncated `\x1b[`, OSC, tab preservation, and DEL.

### Still open from the last round

- **Follow-up issue for the `inspect` vs `diff` numbering mismatch was not filed.** Still reproduces: `inspect` reports alice as `«Email_1»`, `diff` and real `scrub` emit `Mail «Email_2» then «Email_1»`. Fine to fix separately, but README now tells users to run `inspect` then `diff` back to back, so the issue should at least exist before this merges.
- **`readInput(file)` / `addDetectorOptions(cmd)` not extracted.** `diff.ts:275-295`, `inspect.ts:121-141` and `scrub.ts:104-124` are three byte-identical copies of the stdin/file block, and the five detector options are triplicated verbatim.

### Nits

- `LCS_CELL_BUDGET` degrades silently: past 250k cells with unequal lengths, `pairIndexWise` renders a shifted region as a wholesale rewrite with no indication why. A one-line stderr note would fit.
- Side-by-side padding uses `String.length`, so wide chars misalign the columns (`名前 taro@corp.com   | 名前 «Email_1»`), and `wrap()` slices on UTF-16 code units so an emoji on the boundary splits into lone surrogates. Cosmetic.
- Unified-mode `...` separator sits at column 0 while context lines are indented two spaces; `  ...` would line up with the gutter.

will-lamerton (changes requested):
Nice work overall: the `io.ts` extraction is a real cleanup, the `inspect` numbering fix is a genuine bug catch, and the test coverage is solid. Verified locally at `72467fc`: `tsc --noEmit` clean, 27 CLI tests pass, no session written.

Requesting changes on two things.

**1. The LCS-budget fallback produces a wrong diff on realistic input (blocking)**

`diffLines` falls back to `pairIndexWise` when the trimmed middle exceeds `LCS_CELL_BUDGET`. Index-wise pairing is only valid when line counts are equal. `PostalAddressDetector`'s character class includes `\s`, so an address can match across a newline and collapse two lines into one, making the counts unequal.

Repro at PR head: ~800-line file, email on the first line, a multi-line address in the middle, email on the last line:

```
diff: input too large for a full line comparison; showing an index-wise pairing
- start a@x.com
+ start «Email_2»
- body line 1
+ body line «Address_1» please
- body line 2
+ body line 401
...
```

1207 output lines where 6 were correct. Everything after the address is misaligned by one. For a command whose job is "show me exactly what will be sent", that is worse than no output.

Suggestion: skip diffing entirely. You already have `findings` with exact spans, so you can map each span to line numbers in the original and to the corresponding offsets in the scrubbed text. O(n), always exact, and it lets you delete `lcsDiff`, `pairIndexWise` and the budget constant. Failing that, Myers O(ND) is effectively linear here since the real edit count is tiny. At minimum, raise the budget (250k cells trips at ~500 differing lines) and degrade to something honest rather than a fabricated pairing.

**2. `sanitizeLine` silently deletes the OSC payload (blocking)**

`src/cli/sanitize.ts` drops everything from `ESC ]` through BEL/ST, and an unterminated OSC drops the rest of the line. Your own test asserts it:

```ts
t.is(sanitizeLine('\x1b]0;secret@corp.com\x07visible'), 'visible');
```

Content silently vanishing from the display is the exact failure mode this command exists to prevent: a missed detection sitting after an injected `\x1b]` would never show up in the diff. The non-CSI/OSC branch already does the right thing (drops the ESC, prints the rest as literal text). Do the same here, or substitute a visible marker.

Related, non-blocking: only `code === 27` is checked, so the 8-bit CSI introducer 0x9B survives. `sanitizeLine('31mRED-secret@corp.com')` returns the string unchanged. One extra condition for `0x80-0x9f`.

**Smaller items**

- The changeset only mentions `diff`, but `inspect`'s table output changes for any input with 2+ entities of the same category, plus values are now sanitized. Worth its own line.
- `console.error` inside `diffLines` makes the otherwise-pure exported `formatDiff` impure. Return a flag and print at the command layer.
- `formatSideBySide` pads the right column, so every row has trailing whitespace: `"a@b.com            | X                 \n"`. Drop `rPad`.
- `wrap()` and the padding count code points, not display width, so CJK breaks `--side-by-side` alignment. Cosmetic, but this tool handles international text by nature.
- `simulateScrub` uses `new SessionManager()`, which takes the disk-enabled branch and burns a `randomUUID()`. `formatInspectOutput` correctly uses `new SessionManager(undefined, {})`. Make them match so "no session written" is structural.
- The equal-length `pairIndexWise` fast path is correct for this command (scrubbing removes lines but never adds them) but wrong as a general diff, and `formatDiff` is exported and generically tested. `formatDiff('a\nb\nc', 'a\nx\nb')` gives `-b +x -c +b`. Worth a comment stating the assumption.
- `biome check` flags import order in `src/cli/commands/diff.ts:1` and `tests/cli/diff.test.ts:1` (`ava` before `commander`). CI's `biome lint .` skips assist rules so it will not fail the build, but cheap to fix.

**Test gaps**: the budget fallback path, `sanitizeLine` on 0x9B, the `--no-color` / `NO_COLOR` / TTY resolution, `--side-by-side` with unequal del/add counts in one hunk, and the success path of `diff <file>`.

Follow-up idea, out of scope here: the root wart behind item 1's `«Email_2»` is that placeholder allocation is coupled to right-to-left replacement, so the first email in a document is numbered last. Decoupling those would read better, though it would change hashes and session maps.

PR #131, opened 2026-08-31
feat(detectors): add CreditCard, SSN, IBAN, and IP address detectors

### LINKED ISSUE #89: [Feature] New Detectors - Credit Card, SSN, IBAN, IP Address
## Description

Add built-in detectors for high-risk PII: Credit Cards, US Social Security Numbers (SSN), International Bank Account Numbers (IBAN), and IP Addresses.

## Use Case

The current detectors are heavily focused on developer artifacts (API keys, paths). Users migrating enterprise or financial workloads require strict detection of financial and identity data to maintain compliance (PCI-DSS, HIPAA).

## Proposed Solution

- **CreditCardDetector:** Match major card formats and validate using the Luhn algorithm checksum to eliminate false positives.
- **SSNDetector:** Match standard `XXX-XX-XXXX` formats with area/group validation.
- **IBANDetector:** Match international formats with MOD-97 validation.
- **IPAddressDetector:** Match IPv4 and IPv6 structures.

These should be enabled by default.

## Alternatives Considered

Requiring users to write custom rule-packs, but these are so universally required that they belong in the core tool.

## Additional Context

Ciatel-Arlanta:
## Description
Implements the 4 built-in detectors requested in #89 to support high-risk financial and identity PII:
- **`CreditCardDetector`**: Detects 16-digit (Visa, Mastercard, Discover) and 15-digit (American Express in 4-6-5 format) numbers, validated using the Luhn (Mod-10) checksum algorithm to reject barcodes and serial numbers.
- **`SsnDetector`**: Detects US Social Security Numbers (`AAA-GG-SSSS`, spaced, and continuous) with SSA validation rules (Area $\ne 000, 666, 900-999$; Group $\ne 00$; Serial $\ne 0000$).
- **`IbanDetector`**: Detects International Bank Account Numbers with ISO 7064 MOD-97 checksum validation using native `BigInt`.
- **`IpAddressDetector`**: Detects IPv4 ($0 \dots 255$ bounds) and IPv6 (full 8-group and compressed `::` notation), avoiding false matches on semantic versions and dates.

All 4 detectors are enabled by default, and integrated and registered in required files. I have decided the priority based on what felt good to me, would be open to suggestions on a better priority ordering.

Would love to hear any corrections if anything is wrong as well. Took help of AI agents as well so there might be something wrong still even after my manual review, used it for the parts in algorithms where I was not sure how to integrate them and for writing a few tests

Closes #89

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing Notes
- **Automated Tests**:
  - `pnpm ava tests/detectors/*.test.ts tests/core/scrub.test.ts` &rarr; 163 / 163 tests passed.
  - Unit tests for each detector cover positive matches, negative edge cases, checksum failures, and span indexing.
  - End-to-end scrubbing and round-trip rehydration tests added to `tests/core/scrub.test.ts`.
- **Quality Gates**:
  - `pnpm run test:types` (`tsc --noEmit`) &rarr; 0 type errors.
  - `pnpm run test:lint` (`biome lint .`) &rarr; Clean.
  - `pnpm run test:knip` (`knip`) &rarr; 0 unused exports / 0 dead code.
  - `pnpm run test:audit` &rarr; Clean

## Checklist


will-lamerton (changes requested):
Thanks for this — the structure is right. Each detector follows the `Detector` interface and the `phone.ts` exec-loop idiom, categories line up with the `${category}Detector` convention `priorityOf` depends on, the priority renumbering preserves existing relative order, and the Luhn / MOD-97 / SSA implementations are correct.

Verified on the branch: `ava` 302/302 pass, `biome lint` and `knip` clean, and ReDoS probes on all four regexes show no blowup (118KB input scans in 1.7ms). Performance is a non-issue.

Two things to fix before merge, plus some smaller notes.

## Blocking

**1. `pnpm run test:format` fails — CI will go red.** `main` is clean, so this PR introduces it:
- `src/detectors/credit-card.ts:70` — trailing blank line after the closing brace
- `src/detectors/ssn.ts:10-16` — the `return (...)` fits on one line under the Biome config

`pnpm run format` fixes both.

**2. `SsnDetector` fires on any bare 9-digit number.** Both separators are optional in `SSN_REGEX`, so any unpunctuated 9-digit run matches. Through the full `scrub()` pipeline:

```
"Order number 123456789 shipped."  ->  "Order number «Ssn_1» shipped."
"Error code 402551234 returned."   ->  "Error code «Ssn_1» returned."
"qty 100 20 3000 units"            ->  matched as SSN
```

The SSA rules filter roughly 12% of random 9-digit values, so they don't help much here. This is on by default at priority 4 — above Email/Url/Path/Phone — so it wins collisions and silently mangles order IDs, error codes, and part numbers. `phone.ts` already documents the right posture for this repo: *"conservative to avoid false positives... Requires structural markers."*

Suggested shape — require a delimiter for the 3-2-4 form, gate the continuous form behind context:

```ts
const SSN_DELIMITED = /(?<!\d)([0-9]{3})[\s-]([0-9]{2})[\s-]([0-9]{4})(?!\d)/g;
const SSN_CONTEXTUAL = /(?:ssn|social security(?:\s+number)?|tax\s*id)\D{0,10}(?<!\d)([0-9]{3})([0-9]{2})([0-9]{4})(?!\d)/gi;
```

If the continuous form stays unconditional, at minimum move `SsnDetector` below `PhoneDetector` in the priority table.

## Non-blocking

**`::` alone matches as IPv6.** `::1?` makes the `1` optional, so `"a :: b"` -> `"a «IpAddress_1» b"`. `std::vector` is correctly blocked by the lookarounds, but space-delimited `::` shows up in prose and reST. Relatedly, `ip-address.ts:39` (`if (value === ':')`) is dead — no alternative can produce a lone colon. Change `::1?` to `::1`, or make the guard `value === '::'`.

**The `does not match semantic version numbers` test passes for incidental reasons.** `1.2.3.4.5` is rejected by the 5-segment lookaround; `v2.0.0.1` is rejected because `\b` fails between `v` and `2` — neither exercises version logic. A space-prefixed 4-component version does match: `"Upgrade from 1.2.3.4 to 2.0.0.0"` -> both scrubbed. The ambiguity is unresolvable by regex (`1.2.3.4` really is a valid IP), so I wouldn't change the behaviour — but please rename the test so it doesn't imply coverage that isn't there.

**IBAN country codes.** MOD-97 alone gives ~1.1% false positives on random 22-char uppercase tokens (measured 215/20000). Only ~78 country codes issue IBANs; checking `cleaned.slice(0,2)` against that set cuts this ~8x and also catches per-country length mismatches. Worth doing since `IbanDetector` sits above `UrlDetector`/`PathDetector`.

**Smaller:**
- `192.168.1.1/24` scrubs to `«IpAddress_1»/24`, leaving the mask dangling — consider absorbing an optional `/\d{1,2}`.
- `ip-address.ts:35` computes `match.index + (match[0].length - value.length)`, but the only non-capturing parts are zero-width lookarounds, so this is always `match.index`. It's copied from `phone.ts` where the group genuinely is narrower.
- `isValidLuhn`'s 13-19 length guard and its `Number.isNaN` check are both unreachable — the patterns fix the length at 15 or 16 and `replace(/\D/g,'')` guarantees digits. Fine as defence, just noting it.
- `!area.startsWith('9')` is the correct SSA rule, but it also means 9xx ITINs are never flagged. Out of scope for #89 — maybe a follow-up issue.

## Docs

The checklist says docs were updated, but no doc file is in the diff. `docs/features/detectors.md` is now stale in two places: the **Built-in Detectors** list (L36-41) omits all four, and the **Priority & Collision System** list (L51-58) still shows the old 8-entry ordering that no longer matches `collision-resolver.ts`. `README.md:17` is worth extending too.

## Tests

Good breadth on positives, checksum failures, and span accuracy — the `text.slice(start, end) === value` assertions are the right pattern. Gaps:

- The new detectors aren't exercised through `disabledDetectors` / `--disable`. I confirmed the name normalisation handles them (`IpAddressDetector` -> `ipaddress`), but nothing pins it.
- No collision test for the new priorities — e.g. an IP inside a URL resolving to `Url` (works today: `http://192.168.1.1:8080/admin` -> `«Url_1»`).
- `tests/cli/rules.test.ts` still only asserts `PhoneDetector`; adding one new name would guard the `BUILT_IN_DETECTORS` registration.
- No negative corpus. Given the surface these four add, a table of realistic non-PII strings (order IDs, versions, timestamps, MAC addresses, git SHAs) asserting zero findings would be the highest-value test here — and would have caught the SSN issue.

## Security

No new risk. All four are pure string functions, no I/O, no dynamic regex construction, no ReDoS exposure. The `BigInt` try/catch in `iban.ts` is belt-and-braces given the regex guarantees alphanumerics, but fine to keep.

The security-relevant point is the inverse of the usual one: the SSN false positives mean this over-scrubs, and over-scrubbing erodes the signal-to-noise that makes the `inspect`-first workflow viable.

Happy to re-review once the formatting and SSN precision are sorted.

Ciatel-Arlanta:
Thank you for the review, I seem to have missed updating the docs, I thought I had.
Will do the needed changes.

Ciatel-Arlanta:
**Changes since review**

**Blocking**

1. Formatting. Ran `pnpm run format`. `biome ci` is clean, so CI should be green.
2. `SsnDetector` precision. Adopted the suggested delimited/contextual split:

```javascript
const SSN_DELIMITED = /(?<!\d)(([0-9]{3})[\s-]([0-9]{2})[\s-]([0-9]{4}))(?!\d)/g;
const SSN_CONTEXTUAL = /(?:ssn|social security(?:\s+number)?|tax\s*id)\D{0,10}(?<!\d)(([0-9]{3})([0-9]{2})([0-9]{4}))(?!\d)/gi;

```

Both are wrapped in an outer capture group so the finding span covers the digits only, excluding the label the contextual pattern consumes.

Results through `scrub()`:

| Input | Before | Now |
| --- | --- | --- |
| Order number 123456789 shipped. | «Ssn_1» | untouched |
| Error code 402551234 returned. | «Ssn_1» | untouched |
| qty 100 20 3000 units | matched | still matched |
| SSN 123-45-6789 | matched | matched |
| SSN: 219456789 | matched | matched |

The third row is the one open question. `[\s-]` accepts a space, so `100 20 3000` is a well formed delimited run and still matches. Rather than drop it from the corpus, it is pinned by an explicit test that documents the limitation. Changing `[\s-]` to `-` closes it, at the cost of missing unlabelled space separated SSNs. Happy to make that change if you prefer.

`SsnDetector` stays at priority 4, since the continuous form is no longer unconditional.

**Non-blocking**

* IPv6: `::1?` is now `::1`, so a bare `::` in prose no longer matches. The dead `value === ':'` guard is removed, along with the always zero span offset at `ip-address.ts:35`. The IPv6 alternation group is now non capturing since nothing reads it.
* IBAN: added an ISO 13616 registry of the 89 issuing country codes mapped to their exact length, checked before MOD-97. This subsumes the old 15 to 34 length check.
* CIDR: IPv4 absorbs an optional `/0` to `/32` suffix, so `192.168.1.0/24` scrubs whole with no dangling mask.
* Semver test renamed to `does not match 5-segment versions or letter-prefixed versions`, with a comment stating that `1.2.3.4` is a valid IPv4 and stays matched by design. Behaviour unchanged.
* Left as you advised: the `isValidLuhn` length and `NaN` guards, and the 9xx ITIN rule.

**Found while writing the negative corpus**

`4532015000000007x` was scrubbed as a credit card. The `(?!\d)` boundary allowed an adjacent letter, so build IDs and hashes containing a Luhn valid run could match. Boundaries are now `(?<![0-9A-Za-z])` and `(?![0-9A-Za-z])`.

**Docs**

* `docs/features/detectors.md`: all four detectors added to the Built-in list with their validation rules, and the stale 8 entry priority list replaced with the current 12 from `collision-resolver.ts`.
* `README.md`: lines 16 and 23 extended with credit cards, IBANs, SSNs, and IP addresses.

**Tests**

* Negative corpus of realistic non PII strings: order IDs, error codes, versions, timestamps, MAC addresses, git SHAs, tracking numbers.
* `disabledDetectors` coverage for all four, including suffix free names.
* CLI `--disable` coverage with a comma separated list mixing suffixed and bare names.
* Collision test: `[http://192.168.1.1:8080/admin](http://192.168.1.1:8080/admin)` resolves to a single `Url` finding.
* `tests/cli/rules.test.ts` now asserts all four names, guarding `BUILT_IN_DETECTORS` registration.
* Regression tests for the bare 9 digit, bare `::`, embedded card, CIDR, and IBAN country code cases.

**Verification**

209 detector and core tests pass, `tsc --noEmit`, `biome ci`, and `knip` are clean. Diff is 13 files, +200 / -51.

Two pre-existing failures on my Windows machine, `tests/core/rule-packs.test.ts` and `tests/storage.test.ts`, plus a hang in the CLI stdin tests. I confirmed all three behave identically on a stashed baseline, so they are unrelated to this branch.

will-lamerton (changes requested):
Nice work - checksum-gating (Luhn, MOD-97) plus the IBAN country registry is the right approach, and the shared negative corpus is a genuinely useful addition.

Verified on the branch: 325/325 ava, `tsc --noEmit` clean, `biome lint` clean. I also stress-tested the new regexes for backtracking (IPv6 at 8k chars ~0.03ms, IBAN at 4k chars ~0.002ms) - linear, no ReDoS.

Three things I reproduced by running the code that I think need fixing before merge.

## 1. IPv4-mapped IPv6 leaks a partial IP into the output

```js
scrub({ content: 'mapped ::ffff:192.0.2.1 here' })
// -> "mapped «IpAddress_1».0.2.1 here"
```

`IPV6_REGEX` has no IPv4-embedded branch, so it matches `::ffff:192`, while `IPV4_REGEX` separately matches `192.0.2.1`. Two overlapping findings from the same detector reach `resolveCollisions`, which tie-breaks on span length - the 10-char garbage value wins and `.0.2.1` survives into the text sent to the provider.

This is the form nginx, Java `InetAddress`, and Docker emit for dual-stack client IPs, so it is not a corner case, and it is the worst failure mode for this project: partial disclosure that looks scrubbed.

- Add an IPv4-embedded branch to `IPV6_REGEX` (`::ffff:d.d.d.d` and `(?:hex:){6}d.d.d.d`).
- De-overlap inside the detector instead of relying on the resolver's length tie-break. `UrlDetector` already does this with its `alreadyCovered` check - scan IPv6 first, then drop IPv4 matches starting inside an accepted IPv6 span.

## 2. IBAN silently missed when followed by an uppercase word

`src/detectors/iban.ts` - the guard `(?![ ]?[0-9A-Z])` cannot tell "token continues" from "uppercase word follows":

```
'Wire to GB82 WEST 1234 5698 7654 32 today'   -> ["GB82 WEST 1234 5698 7654 32"]
'Wire to GB82 WEST 1234 5698 7654 32 NOW'     -> []      // missed
'acct DE89370400440532013000 ASAP'            -> []      // missed
```

Backtracking yields shorter candidates that fail the length check, so the finding is dropped entirely. "IBAN ... PLEASE PAY" is ordinary payment-email phrasing and the miss is silent.

Narrowing the guard to a real IBAN chunk recovers most of it and keeps every existing IBAN test green (I ran your full corpus against it):

```js
(?![ ][0-9A-Z]{4})(?![0-9A-Z])
```

Still rejects `Ref GB82WEST123456987654321` and `... 32 1111 2222 3333 4444`. Residual gap: a following all-caps word of exactly 4+ letters (`ASAP`) stays blocked - worth a comment pinning that, or a match-generously-then-validate-by-length approach if you want it fully closed.

## 3. SSN fires on ordinary numeric prose

`[\s-]` accepts space, and `SsnDetector` sits at priority 4, above Email/Url/Path/Phone:

```js
scrub({ content: 'Latency was 100 20 3000 ms across runs' })
// -> "Latency was «Ssn_1» ms across runs"
```

You pin this as a known limitation, but I'd push back: the reasoning that led you to gate the *continuous* form behind a label applies with more force here. An unlabelled space-separated 3-2-4 run is overwhelmingly a table row or a measurement, and real SSNs are written with hyphens. Mangling ordinary numbers in an on-by-default, high-priority detector is the worse trade.

Hyphen-only unlabelled, delimiter-tolerant labelled:

```js
const SSN_DELIMITED = /(?<!\d)(([0-9]{3})-([0-9]{2})-([0-9]{4}))(?!\d)/g;
const SSN_CONTEXTUAL =
  /(?:ssn|social security(?:\s+number)?|tax\s*id)\D{0,10}?(?<![\d-])(([0-9]{3})[ -]?([0-9]{2})[ -]?([0-9]{4}))(?!\d)/gi;
```

Ran your whole SSN corpus against this: all positives still pass (`'Tax ID: 456 78 9012 on record'` now via the label), and the prose cases drop to zero. One follow-on - with a delimiter-tolerant contextual pattern, `'Employee SSN is 123-45-6789.'` matches both patterns at the same span, so add a span-dedupe in `detect()`.

Two smaller instances of the same looseness: `\s` includes newline, so `'123\n45\n6789'` matches (use `[ -]`); and mixed separators match (`'123-45 6789'`) - a backreference `([ -])...\1` would tighten it.

## Worth addressing, not blocking

- **Four-part versions scrub as IPv4**: `'version 1.2.3.4'`, `'AssemblyVersion 1.0.0.1'`, `'bumped 2.10.3.1'`. Undecidable by shape, but it is user-visible behaviour in an on-by-default detector and belongs in `docs/features/detectors.md`, not only a test comment. Cheap mitigation: skip a match preceded by `version|v|release|build|assembly`.
- **Consider excluding non-identifying IPv4 ranges**: loopback, `0.0.0.0`, broadcast, link-local (169.254/16), and the RFC 5737 doc ranges identify nobody, and scrubbing them degrades exactly the dev/debug prompts where they appear. Keep RFC 1918 scrubbed - internal topology is worth hiding.
- **`docs/features/threat-model.md:14`** still reads "Email, phone, postal address, path, and URL detectors". Every other listing site was updated.

## Nits

- `iban.ts` `toUpperCase()` is unreachable - the regex only matches `[A-Z]`, so lowercase IBANs are never candidates. Drop it, or add the `i` flag and accept them.
- IBAN registry is missing `FK18` (Falkland Islands, registered 2024). The rest checks out against ISO 13616.
- `credit-card.ts` length guard `13..19` is unreachable; the patterns only produce 15 or 16 digits.
- `credit-card.ts` `[\s-]?` is per-group, so mixed separators and newline-split runs match. Luhn gates it so impact is low, but `[ -]` is tighter.
- IPv6 zone IDs and brackets (`fe80::1%eth0`, `[2001:db8::1]:8080`) leave the suffix behind. The address itself is fully covered, so this one is fine to leave.

## On priority ordering (you asked)

Sound overall - checksummed financial identifiers above shape-based detectors is right, and `IpAddressDetector` at 8 correctly keeps `http://10.0.0.1/x` resolving to a URL. The one I'd revisit is `SsnDetector` at 4: its pattern is the least discriminating of the four (no checksum, purely structural) yet outranks Email/Url/Path/Phone. With the fix in #3 the current slot is fine; without it, loose pattern plus high priority is exactly what produces `«Ssn_1»` for `100 20 3000`.

## Tests

Good coverage - positives, checksum failures, SSA rules, span-slice assertions, round-trips. The gaps map to the findings: no `::ffff:` case, no IBAN-followed-by-uppercase case, and the negative corpus has no 3-2-4 space-form numeric prose and no 4-part version string, which is why #1-#4 pass today. Adding `'Latency was 100 20 3000 ms across runs'` and `'Upgrade to 1.2.3.4'` would pin whichever behaviour you land on.

will-lamerton:
/re-review

Re-running now that both halves of the no-verdict bug are fixed: the rubric no longer asks for the full derivation on span arithmetic (2827acc), and nanocoder continues a turn truncated at the output-token limit instead of reporting it as success (Nano-Collective/nanocoder#1332).

will-lamerton:
/re-review

Third time, with the actual root cause fixed. Every turn was being capped at 4096 output tokens: `@ai-sdk/anthropic` reads the ceiling off the model id and falls back to 4096 for anything that isn't a known Claude model, and nanocoder was setting it under the AI SDK v4 name that v5+ ignores. Reviews ran out of budget mid-analysis and stopped without ever writing a verdict.

- Nano-Collective/nanocoder#1332 - a truncated turn is continued instead of reported as success
- Nano-Collective/nanocoder#1335 - the ceiling is settable again, and provider entries can raise it
- Nano-Collective/.github@78a4e90 - nc-review sets it to 32000

Ciatel-Arlanta:
All three blocking items from the Sep 16 nc-review are addressed:

1. Credit cards (`src/detectors/credit-card.ts`): BIN coverage widened to JCB, Diners (14 and 16 digit), UnionPay, and Maestro, plus 19-digit cards. Luhn stays the gate, separators tightened to space/hyphen only.

2. IPv4 versions (`src/detectors/ip-address.ts`): matches led by a version keyword (`version`, `release`, `build`, `assembly`) are now skipped. A bare `1.2.3.4` with no such keyword is still treated as an IP since the shapes are identical.

3. Spaced SSNs (`src/detectors/ssn.ts`): the delimited pattern accepts a consistent hyphen or space, so an unlabelled `123 45 6789` now matches. The label window is widened to 40 chars. Chose recall over precision here since a miss is a breach.

Tests updated to pin the new behavior, docs updated. 134/134 pass on the touched suites, `tsc` clean. The two `rule-packs`/`storage` failures are pre-existing on Windows (verified on baseline). Ready for another look.

PR #144, opened 2026-09-08
chore: use the shared changelog extractor

will-lamerton:
`scripts/extract-changelog.js` was a per-repo copy. It now lives in [Nano-Collective/.github#4](https://github.com/Nano-Collective/.github/pull/4) and the shared release workflow checks it out, so this copy is dead weight — and worse than dead weight, because it carried two bugs.

## Both were silent

**Section boundaries were found with `(?=\n##+ )`, which also matches `###`.** A changesets changelog opens every version with `### Patch Changes`, so the capture terminated on the *first line of the body*, came back empty, and the script **exited 0 having printed nothing**. The workflow believed it.

**Every release body this repo has published was blank under "What's Changed".** Verified against a real release before changing anything.

**The version was interpolated into a pattern with only `.` escaped**, so `1.0.0(` threw `Invalid regular expression: Unterminated group` — failing a release *after* the package was already on npm. This is what CodeQL flagged as 5 high-severity alerts when the file was copied into nanoterm.

## The shared version

Finds boundaries by walking lines and comparing heading depth, so a section ends at the next heading of the same level or shallower and deeper headings stay in the body. The version is never put into a pattern at all. Verified against this repo's changelog, and against both heading conventions.

Empty output is now a warning in the run log rather than silence — the original bug was invisible precisely because producing nothing looked like success.


will-lamerton:
**Correction to this PR description.** I claimed every release body across every repo was blank under "What's Changed". That was wrong — I reasoned it from the code instead of checking. Verified against each repo's published releases: **only `sentinel` was actually affected.**

The bug is real and this change still stands. It fires whenever a version heading is followed directly by a deeper heading, which is exactly what raw changesets output looks like — and sentinel is the only repo whose `CHANGELOG.md` is raw changesets output. The others were spared by accident, not design: nanocoder post-processes into plain bullets, get-md and prompt-scrubber hand-write theirs without a sub-heading under the version, nanotune links out to the file, and json-up has a different script entirely.

That accident is the actual argument for this PR: any of these repos adopting the changesets default would have started publishing empty release notes with nothing anywhere to say so. The regex-injection half — five high-severity CodeQL alerts — applied to every copy regardless.

Corrected in Nano-Collective/.github#5.

